PR #103, opened 2024-12-03
rst db api

marcellmueller:
- disabled the user api, though leaving in as a nestjs/prisma example plus there is a good chance we can make use of it in the future when we start to build out the admin side
- Add basic recreation_resource table
- create two api routes - one to get all rec resources and one to get one by id
- display a list of all rec sites on landing page
- list links to dynamic url that uses the rec id (still calling `forest_file_id` for now) to fetch a rec resource and display it's information on a page
- some other typescript and setup fixes


mishraomp (approved):
lgtm

PR #106, opened 2024-12-04
chore(docs): update readme to remove qsos sonar badges

mishraomp:


marcellmueller (approved):
Thanks for this! Looks like commitlint needs the `.` removed from the end of the commit message and maybe prettier as well (I enabled it for `.md files since it's supported`)

mishraomp:
> Thanks for this! Looks like commitlint needs the `.` removed from the end of the commit message and maybe prettier as well (I enabled it for `.md files since it's supported`)

fixing it now :)

mishraomp:
now this is happy with me @marcellmueller

PR #128, opened 2024-12-12
fix: autoscaling - scale down policy

mishraomp:
This PR fixes scale down policy which was impacting scale down of tasks


marcellmueller (approved):
Looks good. I was confused about this at first but I read the docs and I think it makes sense now 😅 

Edit: it's early, I thought this was in aws quickstart repo haha

PR #131, opened 2024-12-12
chore: add release workflow

marcellmueller:
Workflow to review test and prod deployments and make a release. Test and prod deployment workflows are just placeholders for now. 

How it works:
- On merge to main we deploy to dev
- After dev is deployed you can manually review deploying to test 
- After deployment to test is successful a release will be created and images will be promoted. When we get more devs it would be good to discuss if we want to make releases on deployment to test or prod, though neither environment is deployed in AWS yet anyways.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -100,7 +100,6 @@ jobs:
   deploy-to-aws-test:
     needs: [release]
     name: Deploys Application to AWS test
-    environment: test
     runs-on: ubuntu-24.04
     steps:
       - name: Deploy to AWS test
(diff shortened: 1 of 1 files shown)


.github/workflows/merge.yml
```diff
@@ -71,9 +42,24 @@ jobs:
       app_env: dev
     secrets: inherit
 
+  deploy-to-aws-test:
+    needs: [deploy-to-aws-dev]
+    name: Deploys Application to AWS test
+    environment: test
```
marcellmueller, line 48: `environment` isn't supported in reusable workflows like `deploy-to-aws-dev` above so trying this out. Likewise, `secrets` isn't supported in the `steps` for this workflow so have to pass the env vars individually.

PR #133, opened 2024-12-15
fix: container dependency , api container should only start when flyway completes successfully

mishraomp:
# Description

This PR fixes container dependency between flyway and Node API and also added flyway  connection retries

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

some documentation links for the related change

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/example_task_definitions.html#example_task_definition-containerdependency

https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ContainerDependency.html

https://documentation.red-gate.com/fd/connect-retries-224919684.html


marcellmueller (approved):
Thanks for this 🙇

PR #134, opened 2024-12-17
chore(cd): infra changes for ecs

mishraomp:
This PR introduces changes how API workloads are scheduled in ECS with fargate and fargate spot.

vars are defined so numbers can be configured differently for different environment.

links:
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/fargate-capacity-providers.html


marcellmueller (approved):
🥇

PR #136, opened 2024-12-19
fix: schemaspy arguments to generate docs and update readme.md

mishraomp:
The database schema documentation will keep up to date via the schedule jobs. 

https://bcgov.github.io/nr-rec-resources/


marcellmueller (approved):
🙇 thanks for this!

PR #145, opened 2024-12-23
Clear dev database before migrations

marcellmueller:
Modified our flyway docker migration image to clean the database before running migrations in dev so we can modify migrations without worrying about breaking the deployment. There are some safeguards and the env `FLYWAY_CLEAN_DISABLED` is set to false be default.


mishraomp (approved):
lgtm

PR #167, opened 2025-01-03
chore: add rst portion of fta database tables to migration file

marcellmueller:
This PR recreates the RST portion of the FTA Oracle database in Postgres.

As discussed this will live in a separate schema called `fta` in our database and we will migrate the CSV dump to this schema. From there we will be able to shape the database as we see fit in our main `rst` schema.


mikevespi (approved):
This is looking good to me, exciting to get the seed data in there once this goes in!

PR #169, opened 2025-01-03
backend search

marcellmueller:
Playing around with making a simple API search route with Prisma ORM filters as well as getting started on the front end search page.

Also got sidetracked updating packages, removing unused ones and moving some unused eslint plugins to the root eslint config 😅


mishraomp (approved):
lgtm

PR #181, opened 2025-01-08
fix: refactor prisma service to make sure there is only one instance

mishraomp:


mishraomp:
I think we can ignore e2e for now

marcellmueller (approved):
Thanks!

PR #186, opened 2025-01-09
chore: remove deploy-pr dependency

marcellmueller:
Fixing a few bugs from the PR refactor:
- Trivy is working correctly now, also had to delete the stale config from the GitHub UI
- Moved Analysis results job so we can set a status check on it
- Removed broken `needs` dependency from `deploy-pr` job since
- Updated some frontend packages


mishraomp (approved):
lgtm

PR #189, opened 2025-01-09
ci: updates and create release workflow

marcellmueller:


.github/workflows/main.yml
```diff
@@ -28,24 +30,73 @@ jobs:
...
     with:
       app_env: dev
       command: apply
       environment_name: dev
       tag: ${{ github.sha }}
     secrets: inherit
-  promote:
+
+  # Separate review job since we can't use GitHub environments in reusable workflows
+  review-test-deployment:
```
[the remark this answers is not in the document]
mishraomp: lgtm, lets roll with it

.github/workflows/.analysis.yml
```diff
@@ -58,7 +58,8 @@ jobs:
     name: Verify Terraform
     uses: ./.github/workflows/.deploy-app.yml
     with:
-      app_env: ${{ github.ref == 'refs/heads/main' && 'dev' || format('ephemeral-{0}', github.event.number) }}
+      # event.number is only available in PRs, so we use github.ref_name as a fallback on main
+      app_env: ${{ format('ephemeral-{0}', github.event.number || github.ref_name ) }}
```
mishraomp, line 62: I am not sure, if app_env could be github.ref_name, since it should be either, dev or test or prod
the only reason we have app_env is we reuse aws env(environment) dev to put both dev and test of the product

marcellmueller: I'm missing something then because how is it working with `ephemeral-pr#`? 

I'm still not entirely sure how I had tf state lock a couple of times when running this on main, so I just wanted to give it a somewhat unique name like `ephemeral-main`.

mishraomp (approved):
lgtm

PR #190, opened 2025-01-09
search updates

marcellmueller:
This PR adds:
- `Clear filters` button for Search page
- Proper pagination - when the user clicks 'Load more' we return the next 10 results and append it to the local data in state
- When the page is reloaded with `?page=2` search params, it will return 20 results instead of the default 10
- Added playwright util to wait for images to be loaded so fix flaky screenshots
- Add `buildQueryString` util and tests
- Add basic tests for search API


mishraomp (approved):
lgtm

PR #210, opened 2025-01-13
chore: ci cancel-in-progress except on main branch

marcellmueller:
To prevent cancelled deployments from multiple merges to main in succession we are conditionally setting `cancel-in-progress: true`.


mishraomp (approved):
lgtm

PR #223, opened 2025-01-15
chore: remove search path from migrations

marcellmueller:
😵‍💫


mishraomp (approved):
that was something very confusing, glad you figured it out, lgtm

PR #227, opened 2025-01-15
feat/fta migration

marcellmueller:
Creating an sql migration which uses AWS `s3_import` plugin to import FTA data in CSV format from s3.

I'm going through table by table in the RDS Query Editor, making any needed modifications to the table and verifying the import works, then updating our FTA schema and saving the statements in a file.

The file is just living in the `/migrations` folder for now, once this is complete we will have to figure out how we are going to run this on a schedule.

- Created an sql migration which uses AWS `s3_input` plugin to import FTA data stored in `csv` format in s3.
- Separate `rst` and `fta` migration containers
- Moved fta database schema migration to the fta container
- `rst` migration will continue to do `flyway clean` though only in dev environment
- `fta` migration will do `flyway clean` in every environment - we will clear the fta schema before re-importing updated data


mishraomp:
ugh, what a pain, but you are going strong...

PR #276, opened 2025-01-23
feat: backend oracle extract and load

mishraomp:
Done
- [x] CI/CD for openshift(GHA)
- [x] Helm Charts for OpenShift
- [x] Secrets For OpenShift - set in github secrets for deployment
- [x] Create IAM User in Rec Resources AWS account to upload to S3 (https://developer.gov.bc.ca/docs/default/component/public-cloud-techdocs/aws/design-build-and-deploy-an-application/iam-user-service/), may be IaC needed.
- [x] Create a secret sync, which will sync secret between AWS SSM and OpenShift Secret and recycle the pods
- [x] Proxy Account for Oracle
- [x] Add Proxy Account Details to GitHub
- [x] IAM automation
- [x] Add namespace and oc token to github secrets (blocked with platform team currently)

PR #281, opened 2025-01-27
feat: adding backend el build deploy for openshift

mishraomp:
This PR is only to add the pipeline and charts, it will be tested later when OpenShift namespace is provisioned and the Proxy Account is created to connect to oracle.

All the fixes and updates will be done later. 

This needs to be merged to main so that manual workflow dispatch action can be triggered and tested.

TODO
- [ ] Add Secrets to Github


marcellmueller (approved):
Awesome work, thanks :)

mishraomp:
merging it to test and makes changes with  https://github.com/bcgov/nr-rec-resources/pull/276

PR #285, opened 2025-01-28
feat/activity filters

marcellmueller:
#324

Add search page filters for activities and set up front end for other filter groups.

- Future filter group issues should theoretically only have to handle the backend querying. The filter menus on the frontend should now automatically handle the rest if given an array with the proper format.
- Filtering activities narrows the search instead of widening it
- Add desktop and mobile filter menus
- We don't have a mobile Figma design so I just did what I thought looked good, worth taking a look at when reviewing
- Filter state is saved in search params in the url
- API takes `activities` search param and accepts a string of ids separated by `_` for multiple activities
- Change string ids for code tables ie `01`, `02` etc to auto incrementing serial integer column ie `1`, `2`. This is better practice and also made the url search params cleaner
- Filtering out the `(Other)` activity type so it's not displayed on the frontend


backend/src/recreation-resource/recreation-resource.service.ts
```diff
@@ -4,69 +4,47 @@ import { PrismaService } from "src/prisma.service";
...
+      },
+      comment: true,
+      status_code: true,
+    },
+  },
+};
+
+type RecreationResourceGetPayload = Prisma.recreation_resourceGetPayload<{
+  select: typeof recreationResourceSelect;
+}>;
```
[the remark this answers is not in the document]
jimmypalelil: lgtm 👍

backend/src/recreation-resource/dto/paginated-recreation-resouce.dto.ts
jimmypalelil, line 1: i just noticed my typo in this file name 

ill fix this soom

backend/src/recreation-resource/dto/paginated-recreation-resouce.dto.ts
```diff
@@ -13,4 +13,11 @@ export class PaginatedRecreationResourceDto {
 
   @ApiProperty()
   limit: number;
+
+  @ApiProperty()
+  activityCount: {
```
jimmypalelil, line 18: not sure what this is for

marcellmueller: Maybe I could name it better but that returns the activities with count for the filters:

Screenshot 2025-01-30 at 11 39 34 AM

frontend/src/components/search/SearchBanner.tsx
```diff
@@ -1,34 +1,50 @@
...
-        }
-      }, 250),
-    [searchParams, setSearchParams],
-  );
+  const handleSearch = () => {
+    if (inputValue === '') {
+      searchParams.delete('filter');
+      setSearchParams(searchParams);
+    } else {
+      searchParams.set('filter', inputValue);
```
hamidhabib-buttons, line 14: (nit): please trim the input values as we dont want any space characters to be included in the input value
`inputValue.trim()`

frontend/src/components/search/filters/FilterGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+        newSearchParams.delete(param);
+      } else {
+        newSearchParams.set(param, updateFilters);
+      }
+      setSearchParams(newSearchParams);
+    }
+  };
+
+  const filtersCount = options?.length + 1;
+  const isShowAllFilters = filtersCount > 5 && !showAllOptions && showMoreBtn;
```
hamidhabib-buttons, line 48: (nit): can we please put the `5` in a constant `const MAX_VISIBLE_OPTIONS = 5`

marcellmueller: Nice suggestion, updated!

hamidhabib-buttons (approved):

jimmypalelil (approved):

PR #287, opened 2025-01-29
update commitlint rules

marcellmueller:
Increase header length and allow capital letters in subject. Apologies to any developers who this annoyed.


mishraomp (approved):
thanks

PR #302, opened 2025-01-31
fix: s3 secret sync

mishraomp:
Fix includes creating the tf deploy ARN into openshift-dev environment in Github
image


marcellmueller (approved):
Nice work, you are making crazy progress thank you.

PR #317, opened 2025-02-03
ci: re-run trivy workflow on failure

marcellmueller:
Trivy sarif-upload is being flaky some days so using the `retry-workflow`. It's not an ideal solution though should mitigate some toil, we will keep an eye on this and perhaps come up with another solution if it still fails.

It should now retry 3 times before failing. Though workflows aren't available through the github cli before merging to main so I was unable to test this completely.


mishraomp (approved):
lgtm

Issue #157, opened 2024-12-31
Display closure notice on Recreation Site and Trails (RST) specific page


BAAlexK:
#### Description:

As a visitor/public, I want to know what rec sites and trails are closed and why so that I can plan my trip.

This ticket is to implement the MVP version of displaying closure notices to the public website. It should look and feel like bc parks closures and use existing data we are pulling from FTA. 

#### Acceptance Criteria:
- [ ] Banner at top of page to make it obvious
- [ ] Matches designs attached below (mimic bc parks closure notices)
- [ ] ...

#### Development Checklist:
- [ ] ...
- [ ] ...
- [ ] ...

#### Dependencies
- Blocked by
- Blocking

#### Notes:
- For MVP use existing closure and alert data/info
- Mockup can be found here: https://www.figma.com/design/TFDtavdpJE2vrB5vEmBB6w/RST---Basic-View?node-id=320-9869&t=Ejnt264WQp2r9cMl-4

JLWade:
@marcellmueller @jaimiebutton  - @BAAlexK  and I found this ticket yesterday and looks like its done?

marcellmueller:
This is mostly done, though I noticed that I introduced a few minor bugs and have a fix incoming.

airliah:
looks good

PR #322, opened 2025-02-04
fix: pagemenu conditional section scrolling issues

marcellmueller:
Follow up for a couple of bugs introduced in #157, I noticed in pages without closures if you scroll down and then scroll back up the top selection in the page menu won't highlight. This is due to the conditional section ref not being in the page.

- added the missing closures menu 🤦 
- refactored the menu array to better handle conditional sections

Screenshot 2025-02-04 at 10 09 28 AM

### DIFF
--- a/frontend/src/components/layout/PageMenu.tsx
+++ b/frontend/src/components/layout/PageMenu.tsx
@@ -4,7 +4,6 @@ type PageSection = {
   sectionIndex: number;
   link: string;
   display: string;
-  visible: boolean;
 };
 
 type PageMenuProps = {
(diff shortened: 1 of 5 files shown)


frontend/src/components/layout/PageMenu.tsx
```diff
@@ -50,7 +50,7 @@ const PageMenu: React.FC<PageMenuProps> = ({
         (section) => section.sectionIndex === Number(index),
       );
       if (selectedSection) {
-        window.location.hash = selectedSection.link;
+        window.location.hash = selectedSection.href;
```
hamidhabib-buttons, line 53: (bug): 

1. We need to use encodeURIComponent for Safety. If href contains special characters i.e. spaces, &, ? it will prevent encoding URL issues.

2. We only need to update the hash if it actually changes. This will improve performance. We can do something like this 

```
if (selectedSection?.href && window.location.hash !== `#${selectedSection.href}`) {
  window.location.hash = encodeURIComponent(selectedSection.href);
}
```

frontend/src/components/rec-resource/RecResourcePage.tsx
```diff
@@ -91,14 +92,58 @@ const RecResourcePage = () => {
...
+    () =>
+      [
+        isClosures ? closuresRef : null,
+        description ? siteDescriptionRef : null,
+        mapLocationRef,
+        campingRef,
+        isThingsToDo ? thingsToDoRef : null,
+        contactRef,
+      ].filter((ref) => !!ref),
+    [description, isClosures, isThingsToDo],
```
hamidhabib-buttons, line 109: (question): `sectionRefs` depends on `closuresRef`, `siteDescriptionRef`, but they are missing from dependencies array. Should that be included?

marcellmueller: Well those sections are just placeholders for now and aren't conditionally rendered, so they aren't dependencies for `useMemo` because there is nothing to change causing `useMemo` to recalculate this.

For example if I remove `isThingsToDo` here I get this eslint warning:
Screenshot 2025-02-07 at 9 10 29 AM

hamidhabib-buttons (approved):

PR #327, opened 2025-02-05
chore: pipeline changes to handle openshift ignores

mishraomp:
This PR separates openshift and aws triggers with path matching.


mishraomp:
unluck, this PR changes both places to trigger both workflows :(

marcellmueller (approved):

PR #334, opened 2025-02-06
fix: move to parallelgc from zgc

mishraomp:
This PR changes the GC used in the app from ZGC to ParallelGC, as the ZGC is more geared towards high amount Heap , more than 100Gigs.


marcellmueller (approved):
🥇

PR #336, opened 2025-02-07
chore: update renovate config to ignore path aws sdk updates

mishraomp:
This PR updates the renovate config to ignore patch updates for AWS SDK , they are very noisy


marcellmueller (approved):
Thanks, you didn't have to do this but I appreciate it! 🙇

PR #342, opened 2025-02-10
fix/migrations

marcellmueller:
Fix for closest_community migration error. Definitely my fault due to rebase error and a mistake I made sorry, need to look into a way to automate testing before merge.

Also tested the test deployment and fixed some issues that revealed themselves. The promote images job failed since it was made before separating the migrations and hasn't been used/updated since then. Also that `postgis` geometry bug came up which I had to manually drop and re-add that extension for the fta deployment to work.

Success workflow deployed to both dev and test:
https://github.com/bcgov/nr-rec-resources/actions/runs/13250117356/job/36986611512

Test is now fully up and running:
https://d1vrz2gt4nmfe5.cloudfront.net/search

### DIFF
--- a/migrations/fta/sql/V1.0.2__fta_to_rst.sql
+++ b/migrations/fta/sql/V1.0.2__fta_to_rst.sql
@@ -14,7 +14,7 @@ select
         when rc.rec_comment_type_code = 'DESC' then rc.project_comment
         else ''
     end as description,
-    rp.closest_community,
+    rp.site_location as closest_community,
     case
         when rp.recreation_view_ind = 'Y' then true
         else false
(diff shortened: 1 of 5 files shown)


migrations/fta/sql/V1.0.2__fta_to_rst.sql
```diff
@@ -48,7 +40,7 @@ select
     forest_file_id,
     case
         when closure_ind = 'Y' then 2 -- Closed
-        when closure_ind = 'N' then 1 -- Open
+        else 1 -- Open
```
marcellmueller, line 43: Sneaking in a fix for many sites not having a status - the way this works in FTA is if a site has never been closed, this is null. If it is reopened then it gets `N` set for `closure_ind`. All sites should be either open or closed.

jimmypalelil (approved):

PR #410, opened 2025-03-03
feat: add rec resource type to the search cards

jimmypalelil:
# Description

Added rec resource type to the search cards in Search Page.

#### Desktop view
image

#### Mobile view
image

### DIFF
--- a/frontend/src/components/rec-resource/card/RecResourceCard.tsx
+++ b/frontend/src/components/rec-resource/card/RecResourceCard.tsx
@@ -20,6 +20,7 @@ const RecResourceCard: React.FC<RecResourceCardProps> = ({
     recreation_activity: activities,
     closest_community,
     recreation_status: { status_code, description: statusDescription },
+    rec_resource_type,
   } = recreationResource;
   const isActivities = activities.length > 0;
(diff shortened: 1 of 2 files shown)


frontend/src/components/rec-resource/card/RecResourceCard.tsx
```diff
@@ -39,7 +40,22 @@ const RecResourceCard: React.FC<RecResourceCardProps> = ({
               />
             </h2>
           </a>
-          <p className="capitalize">{closest_community?.toLowerCase()}</p>
+
+          <div className="d-flex flex-column flex-md-row align-items-md-center align-items-start mb-2 mb-md-0">
+            <span className="fs-6 fw-normal text-dark capitalize">
```
marcellmueller, line 45: I think `text-dark` is a default bootstrap colour which is overriding the global body colour that we have matched with bcparks.ca.

jimmypalelil: i have removed these bootstrap classes

marcellmueller (approved):
:shipit:

PR #411, opened 2025-03-03
feat: add placeholder images in search page cards

jimmypalelil:
# Description

added placeholder images in the search cards for recreation resources that dont have any images

#### Desktop view
image

#### Mobile view
image

### DIFF
--- a/frontend/src/components/RSTSVGLogo/RSTSVGLogo.tsx
+++ b/frontend/src/components/RSTSVGLogo/RSTSVGLogo.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import BCLogo from '@/images/BC_nav_logo.svg';
+import RSTLogo from '@/images/RST_nav_logo.svg';
+import '@/components/RSTSVGLogo/RSTSVGLogo.scss';
+
+export const RSTSVGLogo: React.FC = () => {
+  return (
+    <div
+      className={
+        'rst-svg-logo-container d-flex align-items-center justify-content-center px-4 w-100 h-100'
+      }
+    >
+      <img className={'bc-logo'} src={BCLogo} alt="British Columbia Logo" />
+      <img
+        className={'rst-logo'}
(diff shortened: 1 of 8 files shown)


jimmypalelil:
unit tests will pass once this PR is merged: https://github.com/bcgov/nr-rec-resources/pull/410

frontend/src/components/RSTSVGLogo/RSTSVGLogo.tsx
```diff
@@ -0,0 +1,21 @@
...
+import BCLogo from '@/images/BC_nav_logo.svg';
+import RSTLogo from '@/images/RST_nav_logo.svg';
+import '@/components/RSTSVGLogo/RSTSVGLogo.scss';
+
+export const RSTSVGLogo: React.FC = () => {
+  return (
+    <div
+      className={
+        'rst-svg-logo-container d-flex align-items-center justify-content-center px-4 w-100 h-100'
+      }
```
marcellmueller, line 11:
```suggestion
      className="rst-svg-logo-container d-flex align-items-center justify-content-center px-4 w-100 h-100"
```

Total nit 😅 : we don't need the brackets if there are no dynamic classes. 

For your consideration: why inline so many bootstrap classes when we have a custom `rst-svg-logo-container` to add these into? Why add nested classes for the `.bc-logo`, `.rst-logo` widths instead of inline them as well?

marcellmueller (approved):
This looks great 🥇  Makes the page look so much better now. Added one minor nit comment though this is good to go, thanks!

frontend/src/components/RSTSVGLogo/RSTSVGLogo.tsx
jimmypalelil: - imo using the bootstrap classes is better than writing extra css code

- nested classes were to keep them consistent with the parent where any custom styles are applied through scss files

PR #427, opened 2025-03-07
fix: fta -> rst lowercase error

marcellmueller:


mishraomp (approved):
Oh the case, good catch,  lgtm

PR #430, opened 2025-03-10
feat: enable import and creation of map feature geom table from fta

marcellmueller:


migrations/fta/sql/V1.0.0__init.sql
```diff
@@ -690,7 +690,7 @@ comment on column fta.recreation_map_feature_code.update_timestamp is 'The date
 create table fta.recreation_map_feature_geom (
     rmf_skey serial primary key,
     map_feature_id int,
-    geometry_type_code varchar(4),
+    geometry_type_code varchar(10),
```
jimmypalelil, line 693: were any 10 character long codes?

jimmypalelil (approved):

migrations/fta/sql/V1.0.0__init.sql
marcellmueller: Sorry I think this is a mixup from those columns being flipped, I will revert it thanks

Issue #349, opened 2025-02-11
Display trail and site on map on recreation resource page


BAAlexK:
#### Description:

As a visitor/public, I want to be able to see a trail or site on a map so that I can see where the trail starts and ends on the provincial map and see the boundaries of the site.

[Description of the ticket]

#### Acceptance Criteria:
- [x] See trail linear data on map
     - [ ] BCGW layers are shown for recreation trails 
            - https://catalogue.data.gov.bc.ca/dataset/recreation-line 
- [ ] Need to modify data from oracle
- [x] See site outline data on map

#### Development Checklist:
- [ ] ...
- [ ] ...
- [ ] ...

#### Dependencies
- Blocked by
- Blocking

#### Notes:
- [ ] do we include s58 trails in this ticket?
- [ ] ...

jaimiebutton:
@jimmypalelil: We can use hex code #42814A for the green

BAAlexK:
@airliah looks to be working as expected, with the exception of trails... I think our map shows more!
REC6866 - on old site the interactive map only highlights the first western portion of the trail, our map shows both segments. 100% this will be related to the existing spatial data and database but unclear if this is the business expected behaviour (to only highlight a section when there are disconnected segments)

![Image](https://github.com/user-attachments/assets/8be8f5c1-4966-4902-9490-f0f6c27058a8)

![Image](https://github.com/user-attachments/assets/ea069380-ee12-4660-8801-dff892f0885c)

@jimmypalelil just confirming we are using hex code #42814A for the green?

airliah:
From my perspective we're showing the data based on the latest info we have (same as the GPX data which Bill said is the latest), so I think things look good. But we should ask SME's to review REC6866 in particular and check that it's as expected and if we can add a better explanation to users in future! And the colors are what Jaimie requested and look good to me. I'll leave this open until we chat with them this week, but otherwise should be good to close :)

BAAlexK:
@airliah sounds good!!

PR #433, opened 2025-03-11
feat: added vector maps for rec resource trail and boundary

jimmypalelil:
# Description
Added maps for rec resource trail and boundary

Issue: #349


frontend/src/components/LoadingButton/LoadingButton.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
marcellmueller, line 1: The loading button is a nice touch, I like it. 

Just a couple of things: 

- nit: The loading spinner should either replace the text while loading or there should be a min width so the width of the button doesn't jump back and forth while loading. Buttons should have a fixed width.
- The loading state revealed what I think is a bug. I noticed while checking it with 3g that you can spam the button and it will queue a large amount of queries in a pending state, and slowly complete all of them. I'm not sure the best solution for this, maybe we can cancel in progress in react query?  https://tanstack.com/query/v5/docs/framework/react/guides/query-cancellation or disable the button while loading or both? 

Me spamming the button a bunch of times:
Screenshot 2025-03-14 at 3 15 58 PM

The button stays in loading state until all pending queries complete:
Screenshot 2025-03-14 at 3 16 21 PM

frontend/src/components/search/SearchBanner.tsx
```diff
@@ -26,25 +27,27 @@ const SearchBanner = () => {
...
+                }
+                onKeyDown={handleKeyDown}
+              />
+            </Col>
+            <Col sm={12} md={2} className={'p-md-0'}>
+              <Button className="w-100" onClick={handleSearch}>
+                Search
+              </Button>
+            </Col>
+          </Row>
```
marcellmueller, line 49: There's a small regression in the responsive design

Screenshot 2025-03-14 at 3 37 27 PM

frontend/src/components/search/SearchBanner.tsx
```diff
@@ -26,25 +27,27 @@ const SearchBanner = () => {
...
-            className="btn btn-primary"
-            onClick={handleSearch}
-          >
-            Search
-          </button>
-        </div>
+        <Container className="search-banner-input-container">
+          <Row>
+            <Col sm={12} md={10} className="mb-3 mb-md-0">
+              <input
```
marcellmueller, line 33: Personally I can't stand replacing simple html with abstractions like this or MUI box, especially in an open source project where people might be borrowing our code. Though we should come to a team agreement and stay more consistent rather than a mix match everywhere.

marcellmueller (commented):
Nice work on this, it's working well locally anyways. Just a few comments about things unrelated to the map.

frontend/src/components/LoadingButton/LoadingButton.tsx
jimmypalelil: - I'll fix the width of the button with a fixed width and replace the text with a spinner in the "loading" state.
- we can simply disable the button in "loading" state. that should sovle the issue

frontend/src/components/search/SearchBanner.tsx
jimmypalelil: Since we have react-bootstrap integrated, i thought it'd be best to use these components rather than having duplicated html and css all over the place.

frontend/src/components/search/SearchBanner.tsx
jimmypalelil: good catch.. will fix

frontend/src/components/search/SearchBanner.tsx
marcellmueller: All good just wanted to bring it up as I think it's worth considering not relying on it too heavily as we're only using it due to the bcparks bootstrap theme and a few components bcparks.ca is using react bootstrap for. I'd prefer tailwind over bootstrap css util classes and another UI library if it wasn't for the bcparks theme.

marcellmueller (approved):
This is great. Appreciate the hard work on the maps, splitting up the DTO and other improvements 🙇

PR #444, opened 2025-03-14
chore: change make db migration commands to use flyway

marcellmueller:
As discussed at standup, changed the Makefile to use flyway for migration.Hopefully better catch migration versioning errors earlier. 

Perhaps in the future we can also add a check in ci.


jimmypalelil:
tried it locally and its working as expected

hamidhabib-buttons (approved):

PR #448, opened 2025-03-17
ci: add aws waf web acl to api-gateway infra

marcellmueller:


infrastructure/api/waf.tf
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  visibility_config {
+    cloudwatch_metrics_enabled = true
+    metric_name                = "AppWebACL"
+    sampled_requests_enabled   = true
+  }
+}
+
+resource "aws_wafv2_web_acl_association" "app" {
```
jimmypalelil, line 40: should we not add this at the API GW level instead of the ALB? This will prevent unwanted requests going to the ALB.

marcellmueller: Yeah I initially had added it there, however our api gateway uses HTTP protocol which is [unsupported](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/wafv2_web_acl_association#resource_arn-1) instead of REST which is what I wanted to talk to Om about. 

If you have any thoughts on using REST over HTTP protocol for our api gateway let me know.

jimmypalelil: Oh good to know that its not supported for HTTP API.

I wonder if we could use `CloudFront -> ALB (HTTP) -> ECS` instead. I'm not sure what benefits API Gateway provides if we're just using it as a proxy, though Om might have already explored these options and have better insights.

jimmypalelil: cc: @mishraomp

mishraomp: it is a blocker from OCIO Landing Zone policy, we can expose directly the ALB/ELB, it has to be via API GW. I have not looked at the semantics of HTTP vs REST at API GW, is this something complex? @marcellmueller @jimmypalelil

jimmypalelil: We don’t need to expose ALB. It would still be internal facing. CloudFront distribution should be able to connect to internal facing ALB (🔗 [ref](https://aws.amazon.com/blogs/aws/introducing-amazon-cloudfront-vpc-origins-enhanced-security-and-streamlined-operations-for-your-applications/)


### Regarding HTTP vs REST APIs on API Gateway:

Here’s an excerpt from: [[AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html)](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html)

> REST APIs and HTTP APIs are both RESTful API products. REST APIs support more features than HTTP APIs, while HTTP APIs are designed with minimal features so that they can be offered at a lower price. Choose REST APIs if you need features such as API keys, per-client throttling, request validation, AWS WAF integration, or private API endpoints. Choose HTTP APIs if you don't need the features included with REST APIs.
> 

This basically means that REST API feature would be more expensive for us with WAF integration being the only additional benefit. Thats why I was considering CloudFront as a potentially better option since it would be more cost effective.

mishraomp: > We don’t need to expose ALB. It would still be internal facing. CloudFront distribution should be able to connect to internal facing ALB (🔗 [ref](https://aws.amazon.com/blogs/aws/introducing-amazon-cloudfront-vpc-origins-enhanced-security-and-streamlined-operations-for-your-applications/)
> 
> ### Regarding HTTP vs REST APIs on API Gateway:
> Here’s an excerpt from: [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-vs-rest.html)
> 
> > REST APIs and HTTP APIs are both RESTful API products. REST APIs support more features than HTTP APIs, while HTTP APIs are designed with minimal features so that they can be offered at a lower price. Choose REST APIs if you need features such as API keys, per-client throttling, request validation, AWS WAF integration, or private API endpoints. Choose HTTP APIs if you don't need the features included with REST APIs.
> 
> This basically means that REST API feature would be more expensive for us with WAF integration being the only additional benefit. Thats why I was considering CloudFront as a potentially better option since it would be more cost effective.

this is amazing, looking forward to it :)

PR #496, opened 2025-03-26
feat: add maintenance standard and driving directions

marcellmueller:
- changed `recreation_campsite` since we were just saving a count in a column. Now we are using `recreation_definited_campsite` table and doing a count on the entries in the query itself.
- change `description` on `recreation_resource` to be the non-public description that is on the `fta.recreation_project` table. 
- The public description is in its own table now, `recreation_site_description`. I considered keeping it and driving directions on the main table since they are 1 to 1, however I think it's good practice not to keep large text columns on the main table, and this gives us room to add related columns in the future if we make more complex data related to driving directions etc.
- add maintenance_standard_code to `recreation_resource` table
- Fix bug where facilities still shows in the sidebar menu when the section isn't present
- Some styling updates

## Driving directions:
Screenshot 2025-03-28 at 8 59 40 AM

## Maintenance standard (Maintained):
Screenshot 2025-03-28 at 8 57 31 AM

## Maintenance standard (User maintained):
Screenshot 2025-03-28 at 9 02 27 AM


frontend/src/components/rec-resource/RecResourcePage.tsx
```diff
@@ -282,10 +287,6 @@ const RecResourcePage = () => {
             )}
 
             <Contact ref={contactRef} />
-
-            <p>
-              <a href="/search">Find another Recreation Site or Trail</a>
-            </p>
```
marcellmueller, line 288: This was something I added in early development and has still lingered around somehow even though it's not in the design.

backend/src/recreation-resource/dto/recreation-resource.dto.ts
```diff
@@ -200,13 +182,27 @@ export class RecreationResourceDetailDto extends BaseRecreationResourceDto {
...
+    nullable: true,
+  })
+  driving_directions: string;
+
+  @ApiProperty({
+    description: "The maintenance standard code for the recreation resource",
+    example: "U",
+    nullable: true,
+  })
+  maintenance_standard_code: string;
```
jimmypalelil, line 197: it would be nice to have this as an enum instead of a string. this would make it type-safe and more maintainable.

jimmypalelil (changes requested):

backend/src/recreation-resource/dto/recreation-resource.dto.ts
marcellmueller: Sounds good, I added this. I had tried to do it before but ran into typescript issues in the `formatRecreationResourceDetailResults` function with it. 

I usually try to avoid type assertion as it can cause problems down the line since we are just telling typescript which type it is whether it's true or not. Not sure what you think about that, though I ended up using type assertion since we are using it with other types in this file.

Here is the commit if you want to take a look:
https://github.com/bcgov/nr-rec-resources/pull/496/commits/8ea7c539d06abfda2e5a326cbd4f2738673dffb5

jimmypalelil: Here's the code for the image variants (size code) enum that was defined in the DTO. The enum property in the @ApiProperty decorator is used by openapi cli generator to create the enum types that can be used in the frontend as well.

---

https://github.com/bcgov/nr-rec-resources/blob/43d085ecf2048c565070da8e77595a636538b2a7/backend/src/recreation-resource/dto/recreation-resource-image.dto.ts#L59-L66

marcellmueller: Sorry did you look at the commit I linked above? That's what I did already.

marcellmueller: Though I was asking about type assertion which I usually try to avoid. We are using it quite a bit I noticed when it's supposed to be used sparingly. 

Type assertions are removed by the compiler and won't check the types at runtime.

jimmypalelil (approved):

PR #501, opened 2025-03-27
fix: set default image variant to hpr for rec resources in search results

jimmypalelil:
### Description
Some images do not have 'llc' variant (size_code) generated by DAM and therefore the search cards do not display any images. To fix this we have defaulted to 'hpr' size code as a safe alternative.


backend/src/recreation-resource/recreation-resource.controller.ts
```diff
@@ -53,18 +53,6 @@ export class RecreationResourceController {
     type: String,
     description: "Recreation resource type",
   })
-  @ApiQuery({
```
marcellmueller, line 56: thanks for cleaning up this stuff I missed

marcellmueller (approved):

backend/src/recreation-resource/recreation-resource.controller.ts
jimmypalelil: no worries.. i missed it in the reveiw too

PR #510, opened 2025-03-31
chore: handle recreation reserve on public site

marcellmueller:
This PR handles `Recreation reserves` that are supposed to be displayed on the public site as `Recreation site`. It also creates a source of truth to query the `resource type` since `recreation_map_feature` can have multiple entries, and there is some junk data.

I had initially added a `recreation_resource_type` column to the `recreation_resource` table (#507)  and handled the `Recreation reserve` to `Recreation site` in the fta -> rst import.

Now we are using the view to ascertain the rec resource type on both the search page and the rec resource page, so it's always calculated the same no matter where it's queried from.
  
Thoughts using views:
- It's calculated whenever we query it so always up to date
- We don't have to actually store the type in a column, it's always calculated based on the most recent `recreation_map_feature`
- [Prisma views](https://www.prisma.io/docs/orm/prisma-schema/data-model/views?utm_source=cli&utm_medium=promo-generate-v5-17&utm_campaign=tip-4-nohints) are in preview - I looked into using this for the search refactor, though it doesn't support materialized views even though there are workarounds, and we used raw query anyways to optimize it. 

Thoughts on using a `recreation_resource_type` column (also completed in #507 if you prefer this method):
- Nice to be able to find the resource type on the main table
- Converting and storing a type that isn't the actual type in a column seems like it could cause confusion in the future (Rec reserves with display_on_public_site = true are now Sites)

I'm leaning towards using this view PR instead of the other one, though happy to hear your thoughts.


backend/prisma/schema.prisma
```diff
@@ -1,6 +1,6 @@
 generator client {
   provider        = "prisma-client-js"
-  previewFeatures = ["metrics", "typedSql"]
+  previewFeatures = ["metrics", "views", "typedSql"]
```
jimmypalelil, line 3: it was great to learn about this new feature

jimmypalelil (approved):

Issue #475, opened 2025-03-20
Contact info for beta


airliah:
#### Description:

as a user, I need to be able to contact someone about a rec site/trail if I have issues

![Image](https://github.com/user-attachments/assets/812b7f2a-678c-4dc6-8b0b-f7e9d8159b6d)

#### Acceptance Criteria:
- [x] Open contact form page of original website for now (post beta we'll improve this)
- [x] Point to this page - https://www.sitesandtrailsbc.ca/contact-us.aspx
- [ ] ...

#### Development Checklist:
- [ ] ...
- [ ] ...
- [ ] ...

#### Dependencies
- Blocked by
- Blocking

#### Notes:
- [ ] ...
- [ ] ...

BAAlexK:
@jaimiebutton @airliah are we treating the header contact link the same? 

![Image](https://github.com/user-attachments/assets/66712109-fa9b-4f0f-be5a-735e0ef55203)

BAAlexK:
Perfect, and apologies!! thank you @jaimiebutton

airliah:
@jaimiebutton and/or @harryaolson please advise on whether this should open a new tab (considering it's going to the old site with a different header/navigation)? And consider what we're doing for similar links in the footer in this ticket: https://github.com/bcgov/nr-rec-resources/issues/417. Thanks

BAAlexK:
Decision: open link to Contact Us page - open on same page

PR #518, opened 2025-04-01
chore: add contact url

marcellmueller:
#475


frontend/src/components/rec-resource/section/Contact.tsx
```diff
@@ -20,7 +21,13 @@ const Contact = forwardRef<HTMLElement>((_, ref) => {
               <td>
                 <p>We answer emails weekdays from 9 am to 5 pm Pacific Time.</p>
                 <p>
-                  <a href="">Contact Us</a>
+                  <a
+                    target="_blank"
+                    rel="noopener noreferrer"
```
marcellmueller, line 26: I don't think this is actually necessary as every browser automatically adds it since 2021, leaving it in since we are using it elsewhere and I suppose as safety incase anyone is on a 4 year old browser somehow.

jimmypalelil (approved):

Issue #424, opened 2025-03-05
Add/update page titles to match standards


airliah:
**Describe the task**
We need to follow the same/similar standards as BC Parks for the title tags
- For the **page title tag**, the standard for BC Parks is: 
   - <page title> | BC Parks  OR <protected area name> | BC Parks
- There are some **sub-page standards** as well:
   - Site page (a site is a sub-set of a park, e.g. [Diamond Head](https://bcparks.ca/garibaldi-park/diamond-head-area/))
      - <protected area name> - <site name> | BC Parks
   - Park sub-page (a page with additional info about a specific park, e.g., [Alice Lake Park: Hiking](https://bcparks.ca/alice-lake-park/hiking/))
      - <protected area name: sub page title> | BC Parks

**Acceptance Criteria**
- [ ] Add the correct page and sub-page title tags to all pages (eg. search, rec sites, trails)
- [x] Landing page = Sites and trails BC
- [x] Search = Find a site or trail | Sites and trails BC 
- [x] On Recreation Site or trail page = [Site or trail name] | Sites and trails BC
- [ ] ...

**Additional context**
-

PR #521, opened 2025-04-02
feat: update page titles

marcellmueller:
[#424](https://github.com/bcgov/nr-rec-resources/issues/424)

Updated the page titles and removed the not found component that we returned in RecResourcePage.tsx if a resource wasn't found. This was left over from early development.


frontend/src/routes/constants.ts
```diff
@@ -4,3 +4,12 @@ export const ROUTE_PATHS = {
   REC_RESOURCE: '/resource/:id',
   NOT_FOUND: '*',
 };
+
+export const SITE_TITLE = 'Find a site or trail';
```
jimmypalelil, line 8: isnt the site title supposed to be "Sites and trails BC" ?  

thats what i see in the acceptance criteria in the ticket #424

marcellmueller: Ah yeah good catch thanks, I think I mixed that up in a refactor.

marcellmueller: https://github.com/bcgov/nr-rec-resources/pull/521/commits/05355b6419cd19a65c56d70c73509f309406b433

Updated and removed the non null assertions. Thanks for catching that, would have been embarrassing to go to test 😅

jimmypalelil (approved):

PR #546, opened 2025-04-09
chore: rename maintenance standard section to maintenance type

marcellmueller:
Update the name `Maintenance standard` to `Maintenance type` as well as fix camping anchor tag bug


fbarreta (approved):
LGTM

PR #557, opened 2025-04-09
feat: architecture diagram for RST

mishraomp:
1. Also added ignore for CI when changes are made to docs folder.


marcellmueller (approved):
This looks great, thanks!

mishraomp:
> Thanks for the diagram @mishraomp. it looks great
> 
> I think we should keep the draw.io XML code for the diagram for future iterations. Or we could just add a link to the diagram with edit access for Team Orca if possible.
> 
> Was there a specific reason for removing it?

Yeah, I found, we can use the svg as source as well in the draw io tool, so thought of avoiding unnecessary file storage

Issue #337, opened 2025-02-07
Run E2E tests on dev after deployment


marcellmueller:
**Describe the task**
Currently we are running end to end tests in CI on our app running with docker-compose. This is still useful as it tests our containers and ensures we didn't accidentally introduce any regressions in our updates.

Though we should run E2E tests against our app in AWS so we truly test end to end including our cloud infrastructure. Since we need to deploy our updates we will have to run it after we deploy to dev, but before we deploy to test. 

**Acceptance Criteria**
- [ ] ...
- [ ] ...
- [ ] ...

**Additional context**
- Add any other context about the task here.
- Or here

PR #563, opened 2025-04-11
ci: run e2e tests on dev after deployment

marcellmueller:
#337

### e2e ci changes:
- the e2e workflow can run e2e locally ie docker-compose or using a live base url passed in via environment secrets
- e2e tests now run against the dev environment after dev has been deployed
- e2e tests now run against the test environment after test has been deployed
- I may change this but both dev and test e2e use a new project `rst-dev`. Theoretically they should be the same so I think we can use the same one, but we will see how it goes and I can create another `rst-test` if needed

### Testing changes:
- Tests are rewritten to work against both the dev fixtures and the live data
- Instead of verifying specific results, or the number of results we are verifying that there are results
- Verifying that each search results card contains the text that was searched for
- For filters, we are intercepting the response and validating that the responses include the expected data - this only works for types and activities filters as the others we aren't actually fetching the data for each rec reserve

### DIFF
--- a/.github/workflows/.e2e.yml
+++ b/.github/workflows/.e2e.yml
@@ -1,21 +1,32 @@
-name: Playwright Tests
+name: Playwright e2e tests
 
 on:
   workflow_call:
     inputs:
+      environment:
+        description: 'The GitHub environment to run the tests in'
+        required: false
+        type: string
       tag:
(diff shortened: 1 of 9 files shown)


fbarreta (approved):
LGTM

PR #581, opened 2025-04-14
fix(search-card): adjusted the margin for the rec status display

jimmypalelil:
fix: The rec status (Open/Closed) were positioned too far from the right side. I have removed the right margin and adjusted the total width of the `icon + status` display

image


frontend/src/components/search/SearchPage.tsx
```diff
@@ -112,75 +112,71 @@ const SearchPage = () => {
           </Col>
 
           <Col md={12} lg={9} className="pe-lg-0">
-            <div className="w-100">
```
jimmypalelil, line 115: just removed the extra `div` here

marcellmueller (approved):
Something weird going on in the first Happo diff, I have no idea what is causing it. It doesn't look like it would be anything in this PR, though its working fine in Chrome on my machine.

Other than that LGTM 🥇

jimmypalelil:
@marcellmueller 
I noticed that as well and re-ran the tests assuming that it will replace the screenshots with the new corrected ones but they are still the same.

PR #583, opened 2025-04-14
fix: site status

marcellmueller:
Considered doing a view as a source of truth similar to `recreation_type` but I don't think that would work as unlike the type, the issue with status is that a lot of sites just don't have entries at all. Since prisma doesn't support `coalesce` I'm just handling it when we format the data.

- **fix: display rec resources with no status as open**
- **fix: rec resource card status alignment when no activities**

Also fixed the alignment when there were no activities, the status was on the left of the card:
Screenshot 2025-04-14 at 4 11 40 PM

Fixed:
Screenshot 2025-04-14 at 4 11 50 PM


fbarreta (approved):
LGTM

PR #584, opened 2025-04-15
chore: ui tweaks

fbarreta:
Ui tweaks on:

- font size
- more granular responsiveness


frontend/src/components/layout/Footer.scss
```diff
@@ -90,12 +112,12 @@
     position: relative;
     color: $colorGold;
     border: none;
-    font-size: 12px;
+    font-size: 16px;
```
marcellmueller, line 115: why is this one px and the font size above is rem? maybe just make this `1rem` for consistency.

frontend/src/components/layout/Footer.scss
```diff
@@ -35,14 +42,29 @@
...
+    }
+    @media (max-width: 950px) {
+      width: 130px;
+    }
+    @media (max-width: 900px) {
+      width: 300px;
+    }
+    @media (max-width: 900px) {
+      width: 300px;
+    }
```
marcellmueller, line 60:
```suggestion
    @media (max-width: 900px) {
      width: 300px;
    }
```

This one is duplicated.

Though breakpoints every 50px seems like a lot, all good if you think this is the best way but I wonder if we should use grid/fractions or percentage along with min and max width.

marcellmueller (commented):
Looks good just a couple tiny things for your consideration thanks!

frontend/src/components/layout/Footer.scss
fbarreta: The issue I found with that is that we have 3 columns with dynamic size so in smaller interactions it would break the layout, so that's why I added this strict granularity.
(Also left a duplicate by mistake, that I'm removing ....)

marcellmueller (approved):
Looks good thanks!!

PR #602, opened 2025-04-17
Feat/380 site operator pr

fbarreta:
Implementation of site operator using API request.

- Open API mapping of FSA API
- Call to endpoint to fill site operator
- Design change moving the site operator to the bottom, and omitting it in case it doesn't find the results.


backend/src/recreation-resource/service/fsa-resource.service.ts
```diff
@@ -0,0 +1,22 @@
...
+export class FsaResourceService {
+  constructor(private readonly clientApiService: ClientAPIService) {}
+
+  async findByClientNumber(id: string) {
+    const response = this.clientApiService
+      .findByClientNumber(id, FOREST_CLIENT_API_KEY)
+      .pipe(
+        map((response) => response.data),
+        catchError((err) => {
+          throw err;
```
jimmypalelil, line 17: We shouldn't just rethrow the error here. we should catch and transform into meaningful custom exceptions

backend/src/recreation-resource/recreation-resource.module.ts
```diff
@@ -3,10 +3,16 @@ import { RecreationResourceService } from "src/recreation-resource/service/recre
...
 @Module({
   controllers: [RecreationResourceController],
-  providers: [RecreationResourceService, RecreationResourceSearchService],
-  imports: [PrismaModule],
+  providers: [
+    RecreationResourceService,
+    RecreationResourceSearchService,
+    FsaResourceService,
+  ],
+  imports: [PrismaModule, ApiModule],
```
jimmypalelil, line 16: [nit][non-blocker]
--

ApiModule is too generic of a name. we could name it something like `ForestClientAPIModule`

frontend/e2e/poms/pages/search.ts
```diff
@@ -38,7 +38,9 @@ export class SearchPOM {
     await loadMoreBtn.waitFor({ state: 'visible' });
     const searchCardCount = await this.getRecResourceCardCount();
     await loadMoreBtn.click();
-    await this.page.waitForLoadState('networkidle');
+    // await this.page.waitForResponse((response) => {
```
jimmypalelil, line 41: is this just for testing?

frontend/e2e/poms/pages/search.ts
```diff
@@ -80,7 +82,9 @@ export class SearchPOM {
     closest_community,
     status,
   }: RecResource) {
-    await this.page.waitForLoadState('networkidle');
+    // await this.page.waitForResponse((response) => {
```
jimmypalelil, line 85: same as above

frontend/src/components/rec-resource/section/Contact.tsx
```diff
@@ -1,36 +1,54 @@
...
+                <td>
+                  <p>
+                    We answer emails weekdays from 9 am to 5 pm Pacific Time.
+                  </p>
+                  <p>
+                    <a href={EXTERNAL_LINKS.CONTACT}>Contact Us</a>
+                  </p>
+                </td>
+              </tr>
+              {!error && (
```
jimmypalelil, line 38: what about error handling? should we display some text with retry option after the react-query retries are exhausted?

jimmypalelil (changes requested):

backend/src/recreation-resource/service/fsa-resource.service.ts
```diff
@@ -0,0 +1,22 @@
...
+export class FsaResourceService {
+  constructor(private readonly clientApiService: ClientAPIService) {}
+
+  async findByClientNumber(id: string) {
+    const response = this.clientApiService
+      .findByClientNumber(id, FOREST_CLIENT_API_KEY)
+      .pipe(
+        map((response) => response.data),
+        catchError((err) => {
+          throw new HttpException(err.response.data, err.status);
```
jimmypalelil, line 17: The error message from the Forest client API is bubbled up to the client. After discussion with Felipe, we determined this is acceptable for now since the error message is not directly shown to end users and all 5xx errors are displayed with a custom error message that includes a retry option.

Future improvement: Add metrics to track different dependency response types.

frontend/src/service/queries/recreation-resource/recreationResourceQueries.ts
```diff
@@ -54,6 +56,43 @@ export const useGetRecreationResourceById = ({
...
+          id,
+        });
+      }
+    },
+
+    // Only execute query when ID is provided
+    enabled: !!id,
+
+    // only retry on server errors (5xx)
+    retry: (_failureCount, error) => {
```
jimmypalelil, line 85: [non-blocker]
--
react-query has this functionality baked in by default 

see: https://tanstack.com/query/v4/docs/framework/react/guides/query-retries

jimmypalelil (approved):

PR #634, opened 2025-04-24
chore: enable scheduled owasp zap scan

marcellmueller:


mishraomp (approved):
lgtm, super sonic speed

PR #642, opened 2025-04-25
chore: add no results component button

marcellmueller:
Add a `Go back to the full list` button I missed in the 'no search results' component. Initially I had just added a link back to `/search` though since it was the same page and triggered a full page load it caused some weird flashing. 

I did it the proper way and cleared the search params, making for a smoother ux.

Screenshot 2025-04-25 at 11 42 22 AM


fbarreta (approved):
LGTM

PR #649, opened 2025-04-29
chore: photo gallery and maps matomo tracking

marcellmueller:
Follow up for the rest of the matomo tracking requirements as some of these like the map were a bit of a pain.

Tracking updates for:
- Track if the user searches on the landing page or the search page inputs
- Track filter toggle on/off
- Track map zoom in/out and center buttons
- Track if map is panned with mouse/trackpad or zoom with two finger scroll
- Track search card photo carousel left/right button clicks
- Track rec resource page photo gallery opening


fbarreta (approved):
LGTM

PR #652, opened 2025-04-30
fix(api-metrics): use cloudwatch sdk instead of embedded metrics library

jimmypalelil:
# Description

Replaces the `aws-embedded-metrics` library with the official AWS CloudWatch SDK for metrics collection and reporting.

The previous library required a sidecar container running CloudWatch agent on port 25888, which would be excessive for this project. We can switch to a sidecar strategy later based on traffic patterns.


marcellmueller (approved):
Nice work, this looks great.

PR #683, opened 2025-05-02
chore: add scroll to top component

fbarreta:
- Creation of the move to top component.
- Added to the whole site

image


marcellmueller (approved):
This is great, also thanks for the update!

PR #689, opened 2025-05-05
ci: remove dev from fta migration scheduler

marcellmueller:
- Remove dev from the nightly fta scheduler. Dev will inevitably be broken at some points so I don't want it to impede prod being up to date
- Change unused workflow_call to workflow_dispatch so it can be called manually


mishraomp (approved):
lgtm

PR #692, opened 2025-05-06
chore: control filter focus

fbarreta:
- Controls the focus when the "Show All" link is activated via keyboard
- Focus on the next available checkbox ignoring the disabled
- Interrupts algorithm in case of all subsequent checkboxes are disabled


marcellmueller (approved):
This is working well thanks!

Issue #616, opened 2025-04-22
Location search


airliah:
#### Description:

As a user I want to be able to search for rec sites/trails within an area I'm interested in, so I can decide where I want to go when I'm in that area.

#### Acceptance Criteria:
- [x] Add a second search field to the search page (with the same functionality as BC Parks, so if you can copy/paste their code from their repo please do so to save time)
     - [x] Hint text reads: "Near a community" (note - this is different than BC Parks' "Near a city")
     - [x] Users can type in a location
     - [x] Location can be anywhere in BC - this doesn't need to match the 'closest community' listed in the data
     - [x] A dropdown shows location suggestions when users type (showing text characters that match in bold)
     - [x] Users can select the location to add it to the field 
     - [x] Show text if location entered isn't within BC (screenshot below)
- [x] Update the results to include any within a 50km radius from the location entered 
     - [x] Show users the 50km radius from the location
     - [x] Ensure this new search criteria works with existing filters and typing in a name in the name search field
     - [x] Search results are organized based on proximity to location entered, with closest at the top
- [x] Add this new field to the landing page as well (same functionality)
- [x] Update hint text for existing search field to say "Search by name"

Additional notes:
- [ ] See attached screenshots below from BCP (since we don't have mockups yet) for UI reference to follow
- [ ] We will implement the 'current location' option that BCP has in another ticket later
- [ ] Some documentation on the list of cities BCP used is here: https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/Search-city, check their repo and reach out to devs anytime!

#### Development Checklist:
- [ ] Do we need a new API for it? review link: https://apps.nrs.gov.bc.ca/int/confluence/display/BCPRS/Search-city and chat with team lynx if needed

#### Screenshots from BCP:

![Image](https://github.com/user-attachments/assets/91a075d5-4c1b-4396-8bfa-359d6abdac79)

![Image](https://github.com/user-attachments/assets/489f8377-f00e-46ab-b055-655b2de31f22)

![Image](https://github.com/user-attachments/assets/f2d300ae-4c0f-45e3-a051-8b29630a7e19)

harryaolson:
Question have we considered restrictions on search: 
– Word amount (50 words?)
– Valid characters #@)#)$&&@#()

airliah:
@marcellmueller looks great! just one thing I noticed on the landing page for mobile, the blue box background isn't the right size: 

[...]

PR #694, opened 2025-05-06
feat: location search

marcellmueller:
#616

Since our local dev fixtures are small not every city will return results. `100 Mile House`, 108 Mile Ranch`, `Ainsworth hot springs` are some with results I used test. I also tested this in dev the other day and it's working well.


fbarreta (approved):
LGTM

PR #714, opened 2025-05-08
feat(search-filters): implement dynamic filter aggregation logic

jimmypalelil:
**Implemented Dynamic Filter Counts**

Here's how filters work:

**Starting View**
When you first view the list, each filter displays the total number of items in its **category**. Categories include **district**, **type**, **access**, **activities**, and **facilities**.

**Using Multiple Filters**

- When you select filters from **different categories** (like "North District" and "Hiking"), you'll see only items that match **all** your selections. The filter counts across all categories will reflect the **narrowed-down count** of possible results.
- When you select **multiple options in the same category** (like "Hiking" and "Camping") and if that category is either district, access, or type, you'll see items that match **any** of those options. The filter counts for each option *within this single, active category* remain unchanged, showing the total count from the *initial* state since the **"OR" logic** broadens the possible results within that category.

### DIFF
--- a/backend/src/recreation-resource/service/recreation-resource-search.service.spec.ts
+++ b/backend/src/recreation-resource/service/recreation-resource-search.service.spec.ts
@@ -1,309 +1,254 @@
-import { Test, TestingModule } from "@nestjs/testing";
+import { beforeEach, describe, expect, it, Mock, vi } from "vitest";
+import { RecreationResourceSearchService } from "./recreation-resource-search.service";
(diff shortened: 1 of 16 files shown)


marcellmueller (approved):
Nice work fixing and refactoring this. I tested it quite a bit this morning and couldn't find any issues. Also from my unscientific testing on my local machine it is almost as fast or as fast as before so I see that as a big win 🥇

PR #719, opened 2025-05-09
core: location search sort by closest result

marcellmueller:
Noticed that I missed an acceptance criteria where results are sorted by closest location if searching by location.

I did a test push to dev and verified that the results are sorted by distance where in test currently they are still sorted by name.


fbarreta (approved):
LGTM

Issue #661, opened 2025-05-01
Accessibility-> Search page doesn't retain the search turn when returning back to it from a site page


manuji:
**Describe the bug**
On Find a site or trail page, after a successful search if user clicks on a site page to view the site page and navigate back to the Find a site or trail page, the previous search has been reset. This issue fails the [3.3.7 Redundant Entry (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/redundant-entry) criterion. 

**To Reproduce**
1. Go to 'https://d1vrz2gt4nmfe5.cloudfront.net/search'
2. Search for a keyword (eg: Tree)
3. Click on a site from the search results
4. From the Site page try navigating back to the Find a site or trail page either by breadcrumbs or browser back button
5. Check if the search term and the search results retain on the page. 

**Expected behavior**
When navigating back to the Find a site or trail page, the search term and the results should retail. 

**Screenshots**

https://github.com/user-attachments/assets/024f0139-ff81-4d19-adda-01437bab7bac

**Additional context**
Tested on MacOS/ Chrome.

PR #722, opened 2025-05-12
chore: bread crumb search page link has search history

marcellmueller:
[#661](https://github.com/bcgov/nr-rec-resources/issues/661)

Also refactored the breadcrumbs to be static - we are only using them on the rec resource page so there was no need for them to be dynamic. I was having to add all these checks and conditionals to add the history to the search url and it seemed bug prone and overcomplicated. Let me know if you think I should revert it.

### DIFF
--- a/frontend/src/components/layout/BreadCrumbs.tsx
+++ b/frontend/src/components/layout/BreadCrumbs.tsx
@@ -13,11 +13,11 @@ const BreadCrumbs = () => {
       <a href="/" aria-label="Home">
         <img src={HomeIcon} alt="Home icon" />
       </a>
-      <img src={ChevronRightIcon} alt="chevron" />
+      <img src={ChevronRightIcon} alt="chevron icon" />
       <a href={`/search${lastSearch ? lastSearch : ''}`}>
         Find a site or trail
       </a>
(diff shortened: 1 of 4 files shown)


fbarreta:
I agree with it being static, unless @jimmypalelil has something about it, I'm OK.

jimmypalelil:
i am fine with the static approach as well.

frontend/src/components/layout/BreadCrumbs.tsx
```diff
@@ -1,46 +1,24 @@
...
 
-const BreadCrumbs = ({ customPaths }: BreadCrumbsProps) => {
-  const { pathname } = useLocation();
-  const paths = pathname
-    .split('/')
-    .filter((x) => x)
-    .map((path, index) => {
-      return customPaths?.[index]?.route ?? path;
-    });
+  const lastSearch = sessionStorage.getItem('lastSearch');
```
jimmypalelil, line 9: the query won't persist if the tab/window is closed right? is that the desired approach?

marcellmueller: No it won't which I think is intended. This is how bcparks.ca does it.

jimmypalelil (approved):

Issue #658, opened 2025-05-01
Accessibility-> When Site page displayed horizontally, "On this page" partially blocks the rest of the content from viewport


manuji:
**Describe the bug**
When viewing the site pages on Mobile in horizontal orientation, the "On this page" section get frozen and takes up half of the viewport limiting the space to view rest of the content. This issue fails [1.3.4 Orientation (Level AA)](https://www.w3.org/WAI/WCAG22/Understanding/orientation) criterion.  

**To Reproduce**
1. Go to 'https://d1vrz2gt4nmfe5.cloudfront.net/resource/REC160773' on Mobile device
2. Use the device in horizontal orientation. 
3. Scroll down the page
4. The "On this page" section get frozen and takes up half of the viewport 

**Expected behavior**
Don't lock content to either portrait or landscape presentation.

**Screenshots**

![Image](https://github.com/user-attachments/assets/22d1a0c3-366a-4583-b350-c0df89bbb938)

**Additional context**
Tested on iPhone 16 Pro Max with Safari.

Issue #647, opened 2025-04-28
decrease vertical spacing on mobile


airliah:
**Acceptance Criteria**
- [x] On mobile devices when viewing the rec resource page there is a large gap of white space around the list of sections on the page. we need to reduce this gap so it matches our mockups here: https://www.figma.com/design/TFDtavdpJE2vrB5vEmBB6w/RST---Public-Facing-Website?node-id=1841-49173&t=aHjyOSvWyDcGX7HM-0
- [x] Desktop should not be affected
- [ ] ...

Issue #589, opened 2025-04-15
Search page mobile padding issue


BAAlexK:
> Adding the search page mobile padding issue I noticed this weekend here:
> 
> <img width="322" alt="Image" src="https://github.com/user-attachments/assets/114f8719-3ec6-4776-bdd3-041922832538" />
> 
> How it's supposed to look (I think there are other differences as this is from new updates Harry worked on):
> 
> <img width="251" alt="Image" src="https://github.com/user-attachments/assets/9e1a0328-39b8-4916-a61c-78bf83ff98d4" />
>  

 _Originally posted by @marcellmueller in [#453](https://github.com/bcgov/nr-rec-resources/issues/453#issuecomment-2803123100)_

PR #737, opened 2025-05-13
fix: rec resource page menu and mobile padding

marcellmueller:
Combined a few tiny bug fix prs here, #647, #658, #589

- Reduce mobile page padding from 2rem/32px to 1rem/16px
- Reduce rec resource page menu top and bottom margins on mobile
- Fix bug in medium viewport size/tablet mode where the page menu would take up most of the screen
- Fix some regression I made in the landing page search bar - these only happened in firefox and safari


fbarreta (approved):
LGTM

PR #741, opened 2025-05-13
chore: mobile search image and filter style improvements

marcellmueller:
- Use `object-fit: cover` on the mobile search card images just as we do for desktop
- Sort image variants by landscape first since they look/fit better with our card design which has the images with more width than height. This way the more zoomed in portrait images aren't displayed first.
- Update mobile filter styles to closer align with bcparks.ca

Update to how we handle images in the mobile search cards. 

Some sites currently show portait images like this:
Screenshot 2025-05-14 at 9 29 46 AM

Now we sort images to show landscape first since they aren't so zoomed in:
Screenshot 2025-05-14 at 9 27 50 AM

Then we use `object-fit: cover` to display portrait images just like we do on desktop:
Screenshot 2025-05-14 at 9 29 07 AM

Minor updates filter styles to more closely align with bcparks. Minor differences in colour due to acessibility/contrast.
Screenshot 2025-05-14 at 9 25 23 AM
Bcparks:
Screenshot 2025-05-14 at 9 27 06 AM

### DIFF
--- a/frontend/src/components/rec-resource/card/CardCarousel.tsx
+++ b/frontend/src/components/rec-resource/card/CardCarousel.tsx
@@ -6,6 +6,7 @@ import {
   faCircleChevronLeft,
   faCircleChevronRight,
 } from '@fortawesome/free-solid-svg-icons';
+import '@/components/rec-resource/card/RecResourceCard.scss';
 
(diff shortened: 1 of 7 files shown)


fbarreta (approved):
LGTM

PR #759, opened 2025-05-14
Feat/698 reorganize public

fbarreta:
- Re organized app directories
- fixed the pipeline files


marcellmueller (approved):
Nice work! just gotta add the coverage directory to gitignore thanks!

PR #765, opened 2025-05-15
fix: community search param excluded from breadcrumb history

marcellmueller:
Breadcrumb history now includes `community` param so the new location search input is populated on return. This had to be handled as that param isn't used by used by the api so the type wasn't autogenerated.

Screenshot 2025-05-15 at 10 44 29 AM

Also added `position: relative` to the name input container so the clear input button is inside the input as it previously was:
Screenshot 2025-05-15 at 10 43 16 AM

### DIFF
--- a/frontend/src/components/recreation-search-form/RecreationSearchForm.tsx
+++ b/frontend/src/components/recreation-search-form/RecreationSearchForm.tsx
@@ -62,7 +62,7 @@ export const RecreationSearchForm: FC<RecreationSearchFormProps> = ({
       <Row className="gy-3 gx-0 gx-lg-3">
         <Col md={10} className="pe-md-3 pe-lg-2">
           <InputGroup className="search-input-group">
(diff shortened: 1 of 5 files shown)


frontend/src/components/recreation-search-form/RecreationSearchForm.tsx
```diff
@@ -62,7 +62,7 @@ export const RecreationSearchForm: FC<RecreationSearchFormProps> = ({
       <Row className="gy-3 gx-0 gx-lg-3">
         <Col md={10} className="pe-md-3 pe-lg-2">
           <InputGroup className="search-input-group">
-            <Col>
+            <Col className="position-relative">
```
marcellmueller, line 65: Fix for the clear name input button appearing outside of the input regression

fbarreta (approved):
LGTM

Issue #786, opened 2025-05-21
search field bugs


airliah:
**ONE:**
- from the search page enter a city and also something in the name field that won't return any results
- then click the 'back to full list' text link and see that the city is still showing even if the filters and name field are cleared correctly
- the video below shows this too

**TWO**
- see video below, when you type in something in the name field and then select a city (without selecting search), the results only show the city and not what you entered in the name field

**NOTE**
- make sure the fixes work for the landing page as well

**Video**
https://github.com/user-attachments/assets/9eb11d0e-516c-4e7a-a739-db65ef01476b

airliah:
yay thx!

PR #825, opened 2025-05-23
fix: search input state issues

marcellmueller:
#786 
- Fix `Go back to full list` button not clearing the city input search when there are no search results
- Fix if user types in name search field, then selects a city. Previously it would only search by the city and ignore the name input search, now it searches both name and city.
- Fix landing page url
- Fix screenreaders not letting people know there are no results


fbarreta (approved):
LGTM

PR #865, opened 2025-05-28
fix: fta migration github workflow missing app var

marcellmueller:
Looks like I forgot to pass in the app to the fta migration workflow. 

Manually ran it from the actions tab here to verify this works:
https://github.com/bcgov/nr-rec-resources/actions/runs/15305973210


fbarreta (approved):
LGTM

Issue #871, opened 2025-05-29
Deployment taskId length error


marcellmueller:
**Describe the Bug**
Deployment can be flakey due to a bug with the taskId. This issue has popped up before though we should solve it so it doesn't hold up the automated nightly migrations.

```
│ An error occurred (InvalidParameterException) when calling the
│ DescribeTasks operation: taskId length should be one of [32,36]
│ Flyway task failed with exit code: 
```

PR #877, opened 2025-05-30
fix(ci): invalid taskid deployment error

marcellmueller:
#871

This looks to fix the flakey migrations task by retrying up to 5 times if it doesn't return an arn.


fbarreta (approved):
LGTM

PR #881, opened 2025-06-02
feat(rec-resource-map): add kml and gpx download functionality

jimmypalelil:
# Description
- add kml and gpx download functionality for the rec resource map
- removed zoom animation on the map

# Screenshots

image

---
image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 SHELL := /usr/bin/env bash
 
-include ./backend/.env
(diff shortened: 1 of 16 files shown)


marcellmueller (approved):
Thanks for the updates. Looking at the Figma it just needs to to be blue `#003366` aka `$colorBlue` in our scss variables, I'll approve it though.

PR #892, opened 2025-06-04
Poc/823 implement prisma on admin

fbarreta:
Add multi schema and prisma to admin project


marcellmueller (approved):

jimmypalelil:
LGTM.


@fbarreta I thought instead of multi scheme we were just going to create different DB users/roles and grant permissions accordingly. Just curious to know why we changed it back

Issue #672, opened 2025-05-01
Test environment deployment for Admin project


BAAlexK:
This ticket is for the deployment of the Admin project to Test Environment

- [x] deploy FE and BE of admin
- [x] ensure builds and deploys
- [x] cloudfront for FE and backend API
- [x] separate deployment of public/beta and admin
- [x] Test CI/CD

**Related tickets**
- [ ] Blocked by #676 
- [ ] blocking #858 and #899

BAAlexK:
@marcellmueller this ticket is ready for you to add details/fill in so we can refine as a team

marcellmueller:
Does this need to be in this sprint? I thought we were going to just aim to deploy to dev

BAAlexK:
@marcellmueller @jimmypalelil @fbarreta can you discuss as a group and advise if this is doable in the sprint, and if not then can you let us know if it should be deprioritized? Thanks!

jimmypalelil:
i dont think test is required for this sprint. it can definitely be deprioritized since we have the authorizations and login to be implemented which in my view is higher priority

BAAlexK:
removed from Sprint 8 as per Developer recommendation - lower priority than other admin set up tickets @airliah

PR #924, opened 2025-06-11
ci: fix admin e2e conditional workflow

marcellmueller:
#672 

Small PR for deploy admin to test, had to fix how we skip e2e since it's not set up for admin yet. The entire job was being skipped so the review test deployment job wasn't firing.


fbarreta (approved):
LGTM

PR #941, opened 2025-06-18
chore: test deployment

marcellmueller:
- **chore: add cors to api tf config**
- **chore: test deployment**


infrastructure/api/waf.tf
```diff
@@ -61,23 +112,26 @@ resource "aws_cloudfront_distribution" "api" {
   enabled = true
 
   default_cache_behavior {
```
jimmypalelil, line 114: I do see a difference in the cache policy between dev & test environments. The selected option is

Dev: Cache policy and origin request policy (recommended)
Test: Legacy cache settings

Dev used to be `Legacy cache settings` but it got changed recently. was this a result of the changes in this PR?

jimmypalelil: I noticed that the auth token was not being forwarded to the server running on ECS and the rec suggestion api was returning 401.

I manually edited the cache settings in `Cloudfront distro > edit default behaviour` and made the following changes to make it work

image

Issue #852, opened 2025-05-26
🚨 Zap scan fix - Content Security Policy (CSP) Header Not Set


marcellmueller:
#### Description:
https://www.zaproxy.org/docs/alerts/40040-2/
https://www.zaproxy.org/docs/alerts/10055-6/
https://www.zaproxy.org/docs/alerts/10055-4/

Also needed for admin tool and is a blocker to launching that

#### Acceptance Criteria:
- [x] Fix CSP: Wildcard Directive
- [x] Fix CSP: Misconfiguration
- [x] Re-run zap scan and verify that this is fixed (see dev notes)

#### Dev notes:
The zap scan runs on a schedule in test environment since it's more stable. To check that this is fixed in dev, you can change the environment from test to dev in .github/workflows/scheduled.yml zap scan job. Then manually run from the actions tab for your pr.

BAAlexK:
closing as done

PR #946, opened 2025-06-20
chore: fix csp configuration

marcellmueller:
#852

Add CSP config and tested both public and admin sites to resolve any issues. Ran the owasp zap scan which updated the issues and removed the `Content Security Policy (CSP) Header Not Set` issue.

Now that CSP is enabled we have `script-src: unsafe-inline` warning in the frontend though for now this is some good reasonable defaults.

See Github actions bot comments at the bottom of these issues:
[Frontend zap scan](https://github.com/bcgov/nr-rec-resources/issues/637)
[Backend zap scan](https://github.com/bcgov/nr-rec-resources/issues/636)

### DIFF
--- a/public/backend/src/app.ts
+++ b/public/backend/src/app.ts
@@ -14,16 +14,7 @@ export async function bootstrap() {
     await NestFactory.create<NestExpressApplication>(AppModule, {
       logger: customLogger,
     });
-  app.use(
-    helmet({
-      contentSecurityPolicy: {
-        directives: {
-          defaultSrc: ["'none'"],
-          connectSrc: ["'self'"],
-        },
-      },
-    }),
-  );
+  app.use(helmet());
   app.enableCors();
   app.set("trust proxy", 1);
   app.enableShutdownHooks();
--- a/admin/backend/src/app.ts
+++ b/admin/backend/src/app.ts
@@ -18,16 +18,7 @@ export async function bootstrap() {
       logger: customLogger,
     });
 
-  app.use(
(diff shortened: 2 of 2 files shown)


public/backend/src/app.ts
```diff
@@ -14,7 +14,16 @@ export async function bootstrap() {
     await NestFactory.create<NestExpressApplication>(AppModule, {
       logger: customLogger,
     });
-  app.use(helmet());
+  app.use(
```
jimmypalelil, line 17: Do we need csp for our backend server? we dont serve any html content that is parsed by the browser.

public/frontend/e2e/utils/index.ts
```diff
@@ -21,10 +21,13 @@ export const analyzeAccessibility = async (page: Page) => {
 };
 
 export const waitForImagesToLoad = async (page: Page) => {
-  await page.waitForFunction(() => {
-    const images = Array.from(document.querySelectorAll('img'));
-    return images.every((img) => img.complete);
-  });
+  while (true) {
```
jimmypalelil, line 24: if the goal is to attach a timeout here, then can’t we use the timeout property in [page.waitForFunction](https://playwright.dev/docs/api/class-page#page-wait-for-function)?

public/backend/src/app.ts
marcellmueller: I removed this from both admin and public backend

mishraomp (approved):
lgtm

PR #970, opened 2025-06-26
feat(admin-rec-suggestions): added filters in suggestion result to exclude invalid data

jimmypalelil:
# Description

add validation filters to the suggestion results to exclude invalid data.


fbarreta (approved):
LGTM

Issue #909, opened 2025-06-09
Map clusters - public


airliah:
As a user, I want to see how many sites/trails are in an area when I'm zoomed out, so that I can see the number without the clutter and zoom in to see them separated.

#### Acceptance Criteria:
- [x] Spike to determine how to use open layers clustering and if it works for us
- [x] When zoomed out to see all of BC, see clusters of sites/trails when there are a lot within a close range
- [x] When users click on the cluster it zooms in to see them all separated
- [x] When users zoom in the cluster separates and shows each pin separately
- [x] Mobile and responsive design
- [ ] Figma mockups: 

#### Dev Notes:
- [ ] ...
- [ ] ...

#### Dependencies
- Blocked by
- Blocking

#### Notes:

- See organics info map and chat with Om https://organicsinfo.gov.bc.ca/map
- See BCP map for reference and their use of clusters: https://governmentofbc.maps.arcgis.com/apps/instant/basic/index.html?appid=4b3eb47c25ff45fa9f6bff45e3f2a389
- separate ticket created for locations with 2 pins: https://github.com/bcgov/nr-rec-resources/issues/997

jimmypalelil:
> When users zoom in the cluster separates and shows each pin separately
> 

The above acceptance criteria is sufficiently met in this PR. However, there is an edge case where if two or more locations share the same coordinates, they will still show up as a cluster and will never separate into individual location pins. The proper behavior needs to be refined, but I think this is out of scope for this task and can be handled in a follow-up task/PR.

PS: A [possible fix](https://github.com/bcgov/nr-rec-resources/pull/986#issuecomment-3033989066) is to place all location pins radially at a fixed distance from the shared coordinate as the center.

cc @airliah

airliah:
test via: https://d1vrz2gt4nmfe5.cloudfront.net/search?map-feature=true

PR #986, opened 2025-06-30
feat: map clusters

marcellmueller:
#909
Screenshot 2025-06-30 at 3 22 19 PM


public/frontend/src/components/search/SearchMap/layers/recreationFeatureLayer.ts
```diff
@@ -1,123 +1,152 @@
...
+export const recreationSource = createRecreationFeatureSource();
 
-    const label = getCapitalizedName(name);
-    const key = `label-${label}`;
+export const loadFeaturesForFilteredIds = async (
+  filteredIds: string[],
+  source: VectorSource,
+  projection: any,
+) => {
+  source.clear();
```
marcellmueller, line 88: Took me awhile to figure this out. The features were disappearing while we loaded a new source when the ids changed. Not the end of the world but I thought we were going to need a loading state. 

This clears the current source, then adds new features so they don't disappear until we pass in new ones.

public/frontend/declaration.d.ts
```diff
@@ -1,7 +1,7 @@
-export {};
 declare module '*.module.css';
 declare module 'happo-playwright';
 declare module 'happo.io';
+declare module 'ol-ext/layer/AnimatedCluster';
```
jimmypalelil, line 4: animations look great 👍

marcellmueller:
@jimmypalelil thanks for catching that regression, I fixed the rec resource map page.

As for the clustering [check out this commit](https://github.com/bcgov/nr-rec-resources/pull/986/commits/d97dd32a2bf953fbb8c335f85a815af193f21181), I added `clusterZoomThreshold` option which disables clustering at whatever zoom you give it. I intentionally left the console.log so you can monitor when the clusters are intentionally breaking - at 16 zoom currently.

I think we may still need to handle features that are on the same point or extremely close as this won't stop that overlap from happening.

jimmypalelil (approved):

PR #996, opened 2025-07-07
feat: add recreation feature layer preview popup

marcellmueller:


public/frontend/src/components/search/SearchMap/SearchMap.scss
```diff
@@ -7,6 +12,14 @@
   z-index: 9999;
   overflow: hidden;
   background-color: white;
+
+  .search-map-feature-preview {
+    max-width: 250px;
```
marcellmueller, line 17: This is a temporary fix so some of the cards aren't completely broken on mobile, will be hiding the cards and displaying the mobile drawer in the future.

public/frontend/src/components/search/SearchMap/hooks/useClusteredRecreationFeatureLayer.ts
```diff
@@ -49,9 +50,10 @@ export const useClusteredRecreationFeatureLayer = (
   useEffect(() => {
     if (!map || !layerRef.current || !clusterZoomThreshold) return;
 
-    // Disable clustering based on zoom level
     const view = map.getView();
 
+    // Disable clustering based on zoom level by setting the distance to zero
+    // *NOTE*: This doesn't appear to be working for the edge cases of features with identical coordinates
     const updateClusterDistance = () => {
```
marcellmueller, line 57: As the comment says disabling the clustering by setting distance to 0 is not working for features with identical coordinates. I had tried another method of passing in an unclustered source when it hits the threshold however it caused all the features on the map to flash and felt really buggy.

jimmypalelil (approved):

PR #1004, opened 2025-07-08
Feat/1002 images api

fbarreta:
Admin images backend.


admin/backend/src/resource-images/service/resource-images.service.ts
```diff
@@ -0,0 +1,192 @@
...
+        rec_resource_id,
+      },
+    });
+    if (resource === null) {
+      throw new HttpException("Recreation image not found", 404);
+    }
+    const ref_id = await createResource(caption);
+    await uploadFile(ref_id, file);
+    await addResourceToCollection(ref_id);
+    const files = await getResourcePath(ref_id);
```
jimmypalelil, line 106: We need to add a retry option here for collecting the image URLs for the image variants.

DAM starts creating preview images asynchronously after the image is uploaded. If the preview creation fails, then we would have incomplete image variants to display on the public site.

**Proposed Solution:**

- Check if the returned variants contain the three needed to display on the public site
- If not, retry at least 3 times with some delay until we get all the required variants
- After retry attempts, fail with 500 status and implement proper logging and metrics

> PS: this could be addressed in a future ticket if you want. Please create the ticket and link it here or in the code comment

jimmypalelil (changes requested):

admin/backend/src/resource-images/service/resource-images.service.ts
fbarreta: This is going to be addressed on ticket:
https://github.com/bcgov/nr-rec-resources/issues/1013

jimmypalelil (approved):

Issue #536, opened 2025-04-07
Tablet - Rec Site Thumbnail spacing


harryaolson:
**Describe the Bug**
There is no bottom padding in an individual **Rec site or trail listing**  in a list of sites. 

This is happening at: 

- Between 575-683px wide 
- 768px wide 

**Expected Behaviour**
Have equal spacing around content of a **Rec site or trail listing**

**Actual Behaviour**
A few options can happen here. 

Move "Recreation Site" on to the same line as place (In this instance "Merritt)

**Screenshots**
 
![Image](https://github.com/user-attachments/assets/37da8874-b7e6-45a2-a0d9-1d44e53c4b37)
![Image](https://github.com/user-attachments/assets/b4dd88dc-a673-454b-90f7-6f33201a82b9)

BAAlexK:
Needs design

harryaolson:
Mock up on a smaller size https://www.figma.com/design/TFDtavdpJE2vrB5vEmBB6w/RST---Public-Facing-Website?node-id=1552-80334&t=Cwz2XSaNxFS5AQMp-1

harryaolson:
@marcellmueller  here is the card component reference https://www.figma.com/design/TFDtavdpJE2vrB5vEmBB6w/RST---Public-Facing-Website?node-id=1867-32132&t=8gY09bAvPgNfS7Te-1

airliah:
@marcellmueller I think this is unblocked now?

marcellmueller:
@harryaolson do you have a link to designs for this that take into account longer names and/or lots of activity icons?

harryaolson:
@marcellmueller  I've compiled everything in one central place here:
https://www.figma.com/design/TFDtavdpJE2vrB5vEmBB6w/RST---Public-Facing-Website?node-id=2987-44927&t=otp68xIPFxVXrXIH-1

PR #1020, opened 2025-07-10
fix: search card breakpoints

marcellmueller:
#536

- Fix search cards styling in between small and medium breakpoints
- Add `see all` activities link which displays if there are over 4 activities
- Only show max of 4 activities except in between small and medium breakpoints where we show only 3

Screenshot 2025-07-10 at 12 35 15 PM


fbarreta (approved):
LGTM

PR #1042, opened 2025-07-17
chore: add new district codes to database

marcellmueller:


public/frontend/e2e/poms/components/filter.ts
```diff
@@ -62,32 +62,40 @@ export class FilterPOM {
     await filterGroup.getByRole('button', { name: /Show all/ }).click();
   }
 
-  async toggleFilterOn(filterGroup: Locator, filterOption: string) {
-    const checkbox = filterGroup.getByRole('checkbox', { name: filterOption });
+  async toggleFilterOn(filterGroup: Locator, filterPrefix: string) {
```
marcellmueller, line 65: Can't do exact matching due to the counts in the filter label, so checking the prefix now since matching was broken due to multiple Okanagen filters ie `Okanagen`, `Boundary-South Okanagan`, `North Okanagan`.

fbarreta (approved):
LGTM

PR #1092, opened 2025-07-25
chore: add retry to get img resources

fbarreta:
add retry to get img resources

### DIFF
--- a/admin/backend/src/resource-images/service/resource-images.service.ts
+++ b/admin/backend/src/resource-images/service/resource-images.service.ts
@@ -103,7 +103,15 @@ export class ResourceImagesService {
     const ref_id = await createResource(caption);
     await uploadFile(ref_id, file);
     await addResourceToCollection(ref_id);
-    const files = await getResourcePath(ref_id);
+    let files = await getResourcePath(ref_id);
+    if (!this.checkFileTypes(files)) {
+      try {
+        files = await this.retryGetResourcePath(ref_id);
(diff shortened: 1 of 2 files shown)


admin/backend/src/resource-images/service/resource-images.service.ts
```diff
@@ -189,4 +197,25 @@ export class ResourceImagesService {
...
+        (f: any) =>
+          f.size_code === "original" ||
+          f.size_code === "thm" ||
+          f.size_code === "pre",
+      ).length >= 3
+    );
+  }
+
+  private async retryGetResourcePath(ref_id: string) {
+    for (let tries = 0; tries < 3; tries++) {
```
jimmypalelil, line 213: We should ensure that retries are only attempted on retryable errors, and implement **exponential backoff** between attempts. It would also be helpful to add **log statements** to improve observability.

For these options, it’s best to check with **Robert** to see if there’s an SLA around preview generation times. If no SLA exists, we may need to rely on a **trial-and-error approach** to fine-tune the retry logic.

Since we're already using `axios`, integrating `axios-retry` could be a cleaner solution than writing custom retry logic from scratch.

---

**Out of scope but nice to have:**

It would be valuable to **emit metrics** to track the number of retries and the total time taken to gather all links. These metrics will help us optimize retry settings over time.

jimmypalelil (changes requested):

PR #1108, opened 2025-07-30
ci: add cache-control headers during cloudfront s3 sync

marcellmueller:
Separate upload of `index.html` and add `no-cache, must-revalidate` cache control headers to that file to hopefully mitigate the issue Airlia ran into after prod deployment.

using `no-cache` instead of `no-store` as it is less strict - it still allows the browser to temporarily cache but we force the browser to revalidate before using it.

Vite hashes all the asset names so we should be good to cache everything else.

Screenshot 2025-07-30 at 9 04 52 AM

Screenshot 2025-07-30 at 8 57 39 AM


mishraomp (approved):
lgtm

jimmypalelil (approved):

PR #1113, opened 2025-07-30
fix: flakey search page test in ci

marcellmueller:
Fix for flakey `SearchPage.test.tsx` in ci

Screenshot 2025-07-30 at 11 40 28 AM

### DIFF
--- a/.github/workflows/.deploy-app.yml
+++ b/.github/workflows/.deploy-app.yml
@@ -112,7 +112,7 @@ jobs:
         run: |
           echo "VITE_RECREATION_RESOURCE_ASSETS_BASE_URL: $VITE_RECREATION_RESOURCE_ASSETS_BASE_URL"
           BUCKET=$(echo "$S3_BUCKET_ARN" | cut -d: -f6)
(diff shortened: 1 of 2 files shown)


.github/workflows/.deploy-app.yml
```diff
@@ -112,7 +112,7 @@ jobs:
         run: |
           echo "VITE_RECREATION_RESOURCE_ASSETS_BASE_URL: $VITE_RECREATION_RESOURCE_ASSETS_BASE_URL"
           BUCKET=$(echo "$S3_BUCKET_ARN" | cut -d: -f6)
-          cd shared && npm ci
+          cd shared && npm install
```
marcellmueller, line 115: Had to switch to `npm install` which I did in my other pr, however the recent pr to add the cache headers overwrote this. 

What is happening is we have optional dependencies for sass compilers for linux x86 and darwin mac. Due to this our lock files will change depending on where we generate them.

I could remove the darwin one as it's not needed to build in CI, however it's needed for me to build on my machine which is useful for debugging sometimes.

PR #1138, opened 2025-08-07
Feat/373 map disclaimer

fbarreta:
Map disclaimer modal


public/frontend/src/components/rec-resource/RecreationResourceMap/MapDisclaimerModal.tsx
```diff
@@ -0,0 +1,88 @@
...
+      scrollable
+    >
+      <Modal.Body className="map-disclaimer-modal-content">
+        <div className="map-disclaimer-modal-content--header">
+          <nav aria-label="Main header navigation">
+            <div className="navbar-brand">
+              <img
+                src={RSTLogo}
+                alt="Recreation Sites and Trails BC Logo"
+                style={{ height: 37 }}
```
jimmypalelil, line 37: height in scss file would be better

PR #1140, opened 2025-08-08
fix(admin): disable upload form submission if file name is invalid

jimmypalelil:
# Description

disabled upload form submission (enter key press) if file name is invalid


marcellmueller (approved):
This is working well, nice improvement!

PR #1219, opened 2025-09-08
feat(public-map): add link on rec resource map to view the site point on the main search map

jimmypalelil:
### Screenshots
image

---
image

---
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -688,9 +688,9 @@
       "license": "Apache-2.0"
     },
     "node_modules/@bcgov/prp-map": {
-      "version": "0.0.11",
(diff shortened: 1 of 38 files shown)


marcellmueller (approved):
This is great, love the map loading bar. 

Out of scope for this PR but would it be easy to make the loading bar always show when the map is loading? For example if you refresh with the url `http://localhost:3000/search?view=map` you can see the visual noise as the map loads, would be great if it always hid those issues.

jimmypalelil:
> This is great, love the map loading bar.
> 
> Out of scope for this PR but would it be easy to make the loading bar always show when the map is loading? For example if you refresh with the url `http://localhost:3000/search?view=map` you can see the visual noise as the map loads, would be great if it always hid those issues.

yea that would be good ux and i believe there's already a ticket to display a bc map inspired progress bar

PR #1220, opened 2025-09-08
Feat/orca 22 know before section

fbarreta:
Added Know Before You Go session:

image


public/frontend/src/components/rec-resource/RecResourcePage.tsx
```diff
@@ -134,6 +136,9 @@ const RecResourcePage = () => {
     Boolean(spatial_feature_geometry?.length) ||
     Boolean(recreation_resource_docs?.length) ||
     driving_directions;
+  const isReservable =
+    recreation_resource_reservation_info !== null &&
```
jimmypalelil, line 140: nit
--

can be simplified to Boolean(...)

PR #1242, opened 2025-09-16
chore: last fixes for reservations

fbarreta:
Final adjustments after call with @BAAlexK 

- Consider recreation fee (camping fee) as a condition to fees on reservation and Know Before you know sessions
- Change Know Before you know session visibility to only Recreation sites
- Few text adjustments and conditionals for more elegant display (colons, and, spaces).

### DIFF
--- a/public/frontend/src/components/rec-resource/RecResourcePage.tsx
+++ b/public/frontend/src/components/rec-resource/RecResourcePage.tsx
@@ -156,7 +156,7 @@ const RecResourcePage = () => {
       id: SectionIds.KNOW_BEFORE_YOU_GO,
       href: `#${SectionIds.KNOW_BEFORE_YOU_GO}`,
       title: SectionTitles.KNOW_BEFORE_YOU_GO,
-      isVisible: true,
+      isVisible: isRecreationSite,
     },
     {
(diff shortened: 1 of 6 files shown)


marcellmueller (commented):
Hey looks good but can you fix the regression that Alex caught on the landing page?

It's due to the changes in `KnowBeforeYouGo.scss`, the additions to elements like `h3`, `li` etc are affecting other parts of the app due to css cascading in an SPA app. The changes to `h3` caused the style to change a bit on the `Are recreation sites reservable?` green banner.

They should be nested in a class to avoid this.

Maybe nested in a `.know-before-you-go` class so these changes are contained to the component. Thanks :)

```
.know-before-you-go {
  li {
    margin: 10px 0;
    padding: 0;
    color: $colorBlueMed;
  }

  .small-title {
    font-weight: 700;
    font-style: Bold;
    font-size: 16px;
    line-height: 25.6px;
  }

  p.small-title {
    margin-bottom: 6px;
  }

  h3 {
    // Unfortunate use of !important due to bootstrap theme headings using !important
    font-size: 1.5rem !important;
    margin-bottom: 1rem;
  }
}
```

PR #1294, opened 2025-10-03
chore: hide outdated districts

marcellmueller:
Add `is_archived` column to `recreation_district_code` table so we don't have to keep track of retired districts.

The following districts have `is_archived = true`:
- RDQC - Queen Charlotte Islands (Haida Gwaii)
- RDRM - Rocky Mountain
- RDOS - Okanagan
- RDMH - 100 Mile-Chilcotin

Had to update the district_count materialized view to exclude `is_archived`

### DIFF
--- a/public/frontend/e2e/data/filters.ts
+++ b/public/frontend/e2e/data/filters.ts
@@ -1,5 +1,5 @@
 export const districtFilterOptions = [
-  { label: '100 Mile-Chilcotin' },
+  { label: 'Boundary-South Okanagan' },
   { label: 'Cascades' },
   { label: 'Chilliwack' },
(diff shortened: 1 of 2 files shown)


public/backend/src/recreation-resource/constants/service.constants.ts
```diff
@@ -1,5 +1,4 @@
 export const EXCLUDED_ACTIVITY_CODES = [10, 18, 26];
-export const EXCLUDED_RECREATION_DISTRICTS = ['RDQC', 'RDRM'];
```
marcellmueller, line 2: this wasn't being used

fbarreta (approved):

PR #1603, opened 2026-03-25
chore: dam cleanup

fbarreta:
Dam cleanup.


terraform/frontend/public-prod/terragrunt.hcl
```diff
@@ -11,8 +11,8 @@ generate "prod_tfvars" {
   target_env = "prod"
   custom_domains = ["beta.sitesandtrailsbc.ca", "sitesandtrailsbc.ca", "www.sitesandtrailsbc.ca"]
   csp_urls = {
-      image_src = "https://dam.lqc63d-prod.nimbus.cloud.gov.bc.ca https://beta.sitesandtrailsbc.ca https://sitesandtrailsbc.ca"
-      connect_src = "https://dam.lqc63d-prod.nimbus.cloud.gov.bc.ca https://bcparks.api.gov.bc.ca"
+      image_src = "https://sitesandtrailsbc.ca"
```
ayeshmcg: Do we need to have image_src as self as in cloudfront content_Security_policy we do have **self**

ayeshmcg (changes requested):
Nice work on the cleanup. Just a few leftovers:
- DAM env cleanup from e2e
- Remove from docker-compose
- Update admin-architecture-flowchart-animated

Also, just a sanity check -all image icons under assets are currently stored as SVGs, except these newly added ones. Would it be better to store these as SVGs as well, or do we want to keep them as PNGs?

terraform/frontend/public-prod/terragrunt.hcl
fbarreta: Thank you for the comments !

ayeshmcg (approved):

PR #1604, opened 2026-04-02
feat: orca 524 zoom to selected resource

fbarreta:
Zoom to resource when click on suggestion dropdown.


diego-oxd (commented):
Everything looks great to me. 
The only thing is that I don't see the preview card being opened and I think the ticket mentions it. Just want to be sure.

diego-oxd (approved):

PR #1609, opened 2026-04-14
chore: implemented search

fbarreta:
**Description**

A new filter by closest community was added on RecSpace search page.

- Backend changes
-  - Result model changes by adding the closest community attribute
-  - Query modified to add closest community as a search criteria

 
A few screenshots were added bellow, that describes the frontend implementation as well as a test script.

image
Adds a community option on the typeahead.

image
While clicking on the typeahead adds the community filter on the Closest community dropdown and adds a Chip filter.

image
Multiple options can be added via typeahead.

image
Multiple options can also be added via drop down, and clicking "Apply".


admin/backend/src/recreation-resources/recreation-resource.repository.ts
```diff
@@ -228,6 +228,21 @@ export class RecreationResourceRepository {
     };
   }
 
+  private buildClosestCommunityWhere(
+    communities?: string[],
+  ): Prisma.recreation_resourceWhereInput | null {
+    if (!communities) {
```
diego-oxd: I think that if `communities` is an empty array, it wouldn't be caught by this if statement. 
`![]` returns false. I think this might be the reason for the issue I was seeing?

admin/backend/src/recreation-resources/queries/recreation-resource-search.queries.ts
```diff
@@ -163,6 +163,12 @@ export function buildSearchWhereSql(query: AdminSearchQueryDto): Prisma.Sql {
       alias: 'rs',
       column: 'status_code',
     }),
+    buildMultiValueExistsCondition({
+      values: parseIntegerCodes(query.closestCommunity),
```
diego-oxd: I think that unlike other filters, `closestCommunity` would have a value like "PRINCE GEORGE". Which shouldn't be parsed as an integer?

Also, since `closest_community` is part of the same table as the resource, this would end up creating a query with this condition `WHERE rr.rec_resource_id = rr.rec_resource_id`. 

Do we need another helper function?

admin/frontend/src/components/rec-resource-suggestion-form/SuggestionMenu.tsx
```diff
@@ -27,45 +32,75 @@ export const SuggestionMenu = ({
...
+          recreation_resource_type,
+          district_description,
+          name,
+          display_on_public_site,
+        } = option;
+        return (
+          <MenuItem
+            key={rec_resource_id}
+            option={option}
+            position={index}
```
diego-oxd: The list of `communities` and `results` are added as options to the same list under the menu. 
The only issue is that if we use `position={index}` for both and both lists start from 0, this will cause issues for the keyboard navigation. More than one item will have the same index. 
Probably have to start resources from `communities.length`

diego-oxd (commented):
Don't know if it's an issue with my setup, but this branch doesn't work for me locally. 

2 General issues I found:
- The search returns empty
- activity chips are not clearable

https://github.com/user-attachments/assets/41ab49e1-a66b-4099-aa85-4dcb5284a1b9

A couple of other issues I want to ask about and maybe we can all decide together. 
- The public frontend fetches a list of cities/communities from the BC Parks public API and that list is used to show the suggestions in the search/typeahead component. Here we're getting a list of communities based on the available resources in the response. Which means we have to wait until we've typed enough characters to see relevant results and only considers a subset. I think this not a big deal tbh, but would be nice to decide together for a consistent approach either way?
- The other filters seem to use URL-based state instead of react-based state. This seems to be working well and allows sharing of a search URL Just wondering if we should do the same here?

diego-oxd (commented):
https://github.com/user-attachments/assets/1966cf91-0fe3-4bed-84da-368d2f1f77c8

Found another small issue with one of the queries. 
I created a new PR into this branch with the fix. If you agree with that fix, then everything else looks good and this PR should be ready.

fbarreta:
> Screen.Recording.2026-05-06.at.8.28.35.AM.mov 
> Found another small issue with one of the queries. I created a new PR into this branch with the fix. If you agree with that fix, then everything else looks good and this PR should be ready.

Thank you buddy ! you rock !

diego-oxd (approved):

PR #1612, opened 2026-04-15
fix: Standardize time zone management in the frontend

diego-oxd:
Fixes a timezone bug where dates stored in public site were showing up a day off depending on the environment's local timezone. 
  
The root cause: Prisma maps PostgreSQL `date` columns to JS `Date` objects at UTC midnight. Timezone wasn't specified on the formatters. So, a date like `2024-06-30T00:00:00.000Z` would get shifted back to June 29.
                                                                                                                                           
Fix is to default all date formatters to `UTC` so the calendar date is always preserved as-is from the DB. If you actually need to display a datetime in Pacific time (e.g. audit timestamps), you can pass `{ timeZone: 'America/Vancouver' }` explicitly.
                                                                                                                                           
Also added a regression test suite to cover all formatters against a UTC midnight input.                                                 
  
Fixes #ORCA-41                  
## Type of change                                                                                                                                                     
- [x] Bug fix (non-breaking change which fixes an issue)                           
  
# How Has This Been Tested?                                                                                                              
                             
Added unit tests in `dateUtils.test.ts` covering all formatters with a UTC midnight date. Also manually verified dates render correctly
 in the UI.
- [x] UTC midnight dates now format to the correct calendar day regardless of local timezone                                             
- [x] Explicit `America/Vancouver` override still works as expected for real timestamps                                                  
                                                                                                                                     
## Further comments  
Developer notes:
I create/updated the formatting functions so that now we have 2 types of functions.

1. Deals with Date only fields: 
- uses UTC so that the time and timezone information is dropped, we only care about the calendar date)
- used for fields like: `image.created_at`, `now.toISOString()`, `document.created_at`

2. Deals with DateTime/timestamp fields 
- uses the user's browser to determine the timezone and automatically converts to their local time.
- used for fields like `image.date_taken`, `fee.fee_start_date`, `fee.fee_end_date`

Removed `formatGalleryFileDate` and replaced it with `formatDateTimeReadable`. 

The goal is that:
- any time we want to format a datetime/timestamp we'd use either `formatDateTimeReadable` or `formatDateTimeFull`. 
- any time we want to format a calendar date, we'd use either `formatDateReadable`, `formatDateFull` or `formatDateShort`

This should remove all confusion about which formatting function to use and which parameters to pass. Unless there's an edge shouldn't need to pass any argument specifying the timezone. The 2 different types of functions should ensure that calendar dates and timestamp fields are always shown correctly.

Notable exception:
`ignition_date` after doing some research, I found that apparently for this situation it's appropriate to always show the local time of where the fire is happening, not from where the user is. In this case, the timezone will also be shown to avoid any confusion.


fbarreta (approved):
LGTM !

ayeshmcg (changes requested):
@diego-oxd , @fbarreta 

I think it would be better to standardize time zone handling at the application level and handle date/time conversion at the API layer where needed. Right now, the dates stored in FTA are in PST, and during migration they were saved into our database as PST as well, which is why we’re seeing these issues. If we convert timestamps to UTC in the backend, we won’t need to manage these conversions in the frontend.

For dates that still originate from FTA, when sending them in the API response (for example, fee dates), we can apply a specific conversion in the API layer and return them in UTC. That way, we can explicitly handle only the affected fields in the backend instead of introducing scattered fixes across the frontend. When we eventually cut off from FTA, we would just need to update our table columns to timestamp (UTC) and remove these conversions altogether.

**Alternatively**, we could handle the conversion during the migration from FTA to RST and store all these dates in UTC directly. In this case we will have to update the column types in RST from date to timestamp, along with the necessary schema and migration changes. With this approach, even after we cut off from FTA, we wouldn’t need any additional changes for these fields going forward.

What do you guys think? Happy to discuss if you have any questions or concerns.

ayeshmcg (changes requested):
@diego-oxd  
THe simplest fix for this could be whenever we want to display date, we just need to pass as(use timeZone: UTC) , we do not need any more changes, just adding this, fixes our problem(in RecreationFee)
```
{formatDateFull(fee.fee_start_date, { timeZone: 'UTC' })} -{' '}
{formatDateFull(fee.fee_end_date, { timeZone: 'UTC' })}
```

what do you think?

public/frontend/src/components/search-map/preview/WildfireFeaturePreview.tsx
```diff
@@ -101,7 +101,12 @@ const WildfireFeaturePreview: React.FC<WildfireFeaturePreviewProps> = ({
             {ignitionDate && (
               <p>
                 <FontAwesomeIcon icon={faCalendar} className="me-2" />
-                Discovered on {formatDateFull(ignitionDate)}
+                {/* timeZone: Vancouver because ignition times are always reported in Pacific time by BC Wildfire Service */}
```
ayeshmcg, line 104: Small question on the timezone choice here: do we want wildfire dates to always display in BC/Pacific time, regardless of the user’s location?

ayeshmcg (changes requested):
Nice work on splitting formatDate* and formatDateTime*,  that feels like a good direction.

In my opinion, it may be clearer not to set UTC as the default at the shared config level for all formatters. Instead, we could keep UTC only in the date-only formatter path, and use that specifically for cases where we get a date from the backend with no timestamp.
That would make the intent more explicit:
**formatDate** = calendar/date-only values, preserve the day with UTC
**formatDateTime** = real timestamps
I think this separation would reduce future misuse.

diego-oxd:
> Nice work on splitting formatDate* and formatDateTime*, that feels like a good direction.
> 
> In my opinion, it may be clearer not to set UTC as the default at the shared config level for all formatters. Instead, we could keep UTC only in the date-only formatter path, and use that specifically for cases where we get a date from the backend with no timestamp. That would make the intent more explicit: **formatDate** = calendar/date-only values, preserve the day with UTC **formatDateTime** = real timestamps I think this separation would reduce future misuse.

We are doing that. We don't set UTC for all formatters, only for the date-only ones. The datetime formatters will default to the user's local timezone.

public/frontend/src/components/search-map/preview/WildfireFeaturePreview.tsx
diego-oxd: Yeah, to be consistent with the BC wildfire service. But the timezone is also displayed to avoid confusion,

ayeshmcg (approved):

PR #1617, opened 2026-04-21
feat: Show districts images on resource page

diego-oxd:
# Description

Add district images to resource page. 
Just check that the correct image is displayed for each resource. 

On mobile, the image should not be displayed at all.


fbarreta (approved):
LGTM !

PR #1621, opened 2026-04-22
feat(cronjob): add prov_forest_use and org_unit extraction

ayeshmcg:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-856
**Changes:**
- Added new entities:
    - ProvForestUse mapped to THE.PROV_FOREST_USE
    - OrgUnit mapped to THE.ORG_UNIT
- Created repositories:
    - ProvForestUseRepository with support for filtering by fileTypeCode
    - OrgUnitRepository for org unit reference data extraction
- Integrated both datasets into ApplicationService CSV export flow
- Updated extractAndUploadCSVToS3() to include:
    - extractAndUploadProvForestUse()
    - extractAndUploadOrgUnit()
- Ensured both datasets are exported through the same S3 CSV pipeline as existing tables

**Notes:**
prov_forest_use is filtered using fileTypeCode = "F05" for cron job execution

### DIFF
--- a/backend-el/Dockerfile
+++ b/backend-el/Dockerfile
@@ -12,6 +12,7 @@ ARG DEPENDENCY=/app/target/
 COPY --from=build ${DEPENDENCY}/*.jar /app/app.jar
 COPY run_app.sh .
 RUN apt-get update && apt-get install -y --no-install-recommends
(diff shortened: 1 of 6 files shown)


diego-oxd (approved):
Looks good to me, great job! 
I don't know if there's a specific logical reason for the naming but every entity and repo names start with `Recreation`. So, I don't know if we should keep that for these new ones. Either way, not a big deal.

PR #1629, opened 2026-05-01
feat: ORCA-258 before you go section

diego-oxd:
Update "know before you go" section in the resource details page to include 4 new subsections. 
Added the SVGs for the icons. 

This PR includes the changes for both since they are in the same section and have similar tests.
[ORCA-810](https://bcparksdigital.atlassian.net/browse/ORCA-810)
[ORCA-258](https://bcparksdigital.atlassian.net/browse/ORCA-258)


public/frontend/src/components/rec-resource/section/KnowBeforeYouGo.tsx
```diff
@@ -124,6 +131,110 @@ const KnowBeforeYouGo = forwardRef<HTMLElement, KnowBeforeYouGoProps>(
...
+              <p>
+                Before lighting a fire, check current restrictions with{' '}
+                <a
+                  href="https://www2.gov.bc.ca/gov/content/sports-culture/recreation/camping-hiking/sites-trails/alerts#wildfire-info"
+                  target="_blank"
+                  rel="noopener noreferrer"
+                  aria-label="BC Wildfire Service (opens in new window)"
+                >
+                  BC Wildfire Service
+                </a>{' '}
```
ayeshmcg: Hey Diego, nice work! One thing to consider, the row layout is repeated quite a few times. It might be worth extracting it into a separate component(something like `InfoRow`) and reusing that throughout. Happy to chat through it if useful!

ayeshmcg (approved):
Nice work Diego! The component looks good overall. Left one comment inline about some layout duplication worth addressing, but nothing blocking. Good to merge once that's considered!

public/frontend/src/components/rec-resource/section/KnowBeforeYouGo.tsx
diego-oxd: Yeah, great suggestion. 
I added a reusable component for those rows.

PR #1648, opened 2026-05-12
chore: Update texts in know before you go section

diego-oxd:
Update some other text in "know before you go" section to match Figma


ayeshmcg (approved):
LGTM

PR #1649, opened 2026-05-13
fix: ORCA-875 map zoom issue

diego-oxd:


fbarreta (approved):
Looks great. Just minor comments on useSearchInput.ts that would be nice to delete those lines.

PR #1652, opened 2026-05-15
feat: orca 858 typeahead sorting sql

fbarreta:
https://bcparksdigital.atlassian.net/browse/ORCA-858

Sort sql order to attent the request in the following order:

- Names containing the searched string
- Types (Recreation first, then reserves)
- Status (Archived last)


diego-oxd (commented):
I think the first condition in the "order by" makes sense. But then, for the other ones, If we only sort by `rec_resource_type_code` and `rec_status_code` we might get the wrong order. 
Like `RR` (recreation reserve) coming before `IF` (interpretive forest) for `rec_resource_type_code`
for the status_code it might work, but I think it'd be nicer to be more explicit. 
Maybe something like this? What do you think?

```sql
CASE WHEN rrtv.rec_resource_type_code IN ('RR', 'RTR') THEN 1 ELSE 0 END ASC,
CASE WHEN rr.rec_status_code = 'AR' THEN 1 ELSE 0 END ASC,
```

fbarreta:
> I think the first condition in the "order by" makes sense. But then, for the other ones, If we only sort by `rec_resource_type_code` and `rec_status_code` we might get the wrong order. Like `RR` (recreation reserve) coming before `IF` (interpretive forest) for `rec_resource_type_code` for the status_code it might work, but I think it'd be nicer to be more explicit. Maybe something like this? What do you think?
> 
> ```sql
> CASE WHEN rrtv.rec_resource_type_code IN ('RR', 'RTR') THEN 1 ELSE 0 END ASC,
> CASE WHEN rr.rec_status_code = 'AR' THEN 1 ELSE 0 END ASC,
> ```

That makes more sense than relying on the natural order. I'll make this change. Tks !!!

diego-oxd (approved):
Looks good

PR #1653, opened 2026-05-16
feat: Add FDL checkbox and validation

diego-oxd:
## Testing notes
Create mode
- Submit the form with the FDL box unchecked → validation error appears under the checkbox, fee is not saved
- Check the box and submit a valid form → fee is created, log entry created

Edit mode
- Open an existing fee — amount field is disabled by default; fee type is always disabled
- Tick the FDL box -->  amount field becomes editable
- Tick the FDL box without changing any other field -->  save button stays disabled (ticking alone is not enough - there haven't been any changes to save)
- Change the amount, leave FDL unchecked, click Save → validation error under the checkbox, fee is not saved
- Change the amount, tick FDL, click Save → fee is updated, log entry created
- Change a non-amount field (e.g. a day checkbox) without ticking FDL → save proceeds normally, no FDL error
shown

Decided to create a new table to hold the log/audit data. A record in this table is only created when the fee is initially created or when the amount changes. Other changes that don't require the FDL are not recorded. We can still the history of changes through the history table. Choose this approach for the following reasons:
- The history table captures all fee changes. So, finding FDL events means diffing consecutive rows and guessing which changes were FDL related
- If any other code path (migration, admin script, future feature) writes a fee amount without going through the
FDL flow, the implicit audit trail breaks silently with no warning.
- The dedicated table is self-documenting: anyone querying it knows exactly what it contains without needing to understand the history table or doing some SQL queries
- easy to extend and add new fields if necessary. 
- There is a plan to show a clean history of changes (including the user who did it) in the frontend in the future


fbarreta (approved):
Looks great.

PR #1654, opened 2026-05-19
feat: added delete fee feature

ayeshmcg:
**Card:** https://bcparksdigital.atlassian.net/browse/ORCA-424

**Changes:**
- Updated recreation_fee and history tables to add:
   - is_deleted
   - deleted_by
   - deleted_at
- Added Delete Fee API
- Updated unique constraint on recreation_fee to apply only when is_deleted = false
- Updated fee retrieval queries to exclude deleted fees (is_deleted = false)
- Added Delete Fee confirmation modal
- Updated Recreation Resource Fee form modal to show delete icon
- Updated tests

**To Test**
- Go to Recreation Resource: http://localhost:3001/
- Select a recreation resource (e.g. http://localhost:3001/rec-resource/REC204117/files)
- Navigate to the Fees section
- If fees already exist, verify the delete icon appears under Actions
- Test fee deletion:
   - A confirmation modal should appear
   - The fee should be removed from the table
   - The deleted fee should no longer be visible on the public site
- Verify that after deleting a fee, you can add a new fee with the same fee type again


shared/src/styles/variables.scss
```diff
@@ -32,6 +32,7 @@ $colorGreenLightest: #e0eddb;
 $colorGreenDarker: #2e8540;
 $colorHeaderDevEnvironment: #d2d945;
 $colorHeaderTestEnvironment: #c7e3fd;
+$color-grey100: #f1f1f1;
```
diego-oxd: Seems like this is only used for the colour of the text in the confirmation button of the modal. But I think that should just be white.

admin/frontend/src/pages/rec-resource-page/components/RecResourceFeesSection/EditSection/DeleteFeeConfirmationModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+interface DeleteFeeConfirmationModalProps {
+  show: boolean;
+  fee?: RecreationFeeUIModel;
+  isDeleting?: boolean;
+  onCancel: () => void;
+  onConfirm: () => void;
+}
+
+export const DeleteFeeConfirmationModal = ({
```
diego-oxd: I see 2 different designs for the confirmation modal on Figma. But we might want to keep it consistent with the existing modal for deleting images. The cancel and delete next to each other at the bottom right and using outline?

admin/frontend/src/pages/rec-resource-page/components/RecResourceFeesSection/RecResourceFeesTable.tsx
```diff
@@ -18,6 +22,32 @@ export const RecResourceFeesTable = ({
   recResourceId,
 }: RecResourceFeesTableProps) => {
   const { canEdit } = useAuthorizations();
+  const deleteFee = useDeleteFee();
+  const [selectedFee, setSelectedFee] = useState<RecreationFeeUIModel>();
+  const [tableFees, setTableFees] = useState<RecreationFeeUIModel[]>(fees);
```
diego-oxd: It might be cleaner to not create state here with the `fees` prop and then updating manually the state here. 
We can rely on the hooks updating the props and then this would re-render automatically to remove the fee?
I think the "getFees" hook doesn't exist yet, but it be cleaner to use that. We probably need to update how fees are gotten in `RecResourceFeesSection` to use that hook. 
Edit: I added a PR into this branch to show what I mean

admin/frontend/src/pages/rec-resource-page/components/RecResourceFeesSection/EditSection/DeleteFeeConfirmationModal.tsx
ayeshmcg: We will definitely need two delete confirmation modal, one generic delete confirmation modal and and the other one should be delete file

admin/frontend/src/pages/rec-resource-page/components/RecResourceFeesSection/RecResourceFeesTable.tsx
ayeshmcg: Thanks for pointing that out, I have already implemented this hook,  I was thinking to cover in a separate tech debt, will implement in this only

fbarreta (approved):

ayeshmcg:
@diego-oxd  the PR has been updated

PR #1658, opened 2026-05-20
feat: orca 148 display natural resource district

fbarreta:
Added query on the natural_resource_org_unit table and mapped the result to the ResultDto.
Some Dto adjustments.
Added Natural resource district name at overview section.


ayeshmcg (approved):
Nice work @fbarreta , its just missing fixture to populate local db(natural_resource_org_unit), I have added that in https://github.com/bcgov/nr-rec-resources/pull/1659

PR #1662, opened 2026-05-21
chore: remove not visible highlight

fbarreta:
https://bcparksdigital.atlassian.net/browse/ORCA-925

Remove the grey background on the typeahead if the resource is no visible on website.


diego-oxd (approved):
Looks good!

PR #1664, opened 2026-05-22
feat: Add boundaries and lines for resources

diego-oxd:
- Added new hooks to manage the new layers, keeping the pattern consistent with other layers (fires and rec pins)
- The 2 URLs don't return the appropriate CORS headers, so we have to route requests through the backend. A new backend service/endpoint was added to proxy the frontend requests to the ArcGIS service. It doesn't touch the DB, it just acts as a router with some validation to make sure the request is legitimate.
- The ArcGIS services let us fetch resources based on a list of provided IDs. This is ideal because we can request exactly the boundaries/lines we need.
  - The strategy: as soon as the zoom level hits our threshold, we fetch the boundaries/lines for all rec resources visible on screen. If the visible resources change, we refetch.
  - I tried other strategies (fetching upfront, bbox, tile-based), but all of them ended up pulling a lot of data we don't need (inactive resources, `display_on_website: false`, etc.). The ID-list approach also means data that's no longer in view doesn't stay in memory, which keeps the app from slowing down over time.
- Site boundaries and trail lines both appear at zoom 10 (per the ticket)


fbarreta (approved):
That's actually a great work.
Did some debugging to learn about the data flow, and the useMemo approach on the layers was spot on.
I would like to have an opportunity to walk through it with you, so I could undestand better the benefits of it.

diego-oxd:
> That's actually a great work. Did some debugging to learn about the data flow, and the useMemo approach on the layers was spot on. I would like to have an opportunity to walk through it with you, so I could undestand better the benefits of it.

Yeah, we can have a call to talk about it!

PR #1665, opened 2026-05-22
chore: fix capitalization occurrences

fbarreta:
https://bcparksdigital.atlassian.net/browse/ORCA-923
Fix capitalization function so it can capitalize the chars after a "(".
Standardize the use of the function through the sessions.


diego-oxd (approved):
Looks good to me, but I wonder if for the names of resources, we should simply leave them as is. It seems like they are already being stored correctly in the DB?
Either way feel free to merge if not.

PR #1671, opened 2026-05-27
feat: add fee sub type

ayeshmcg:
**Card:** https://bcparksdigital.atlassian.net/browse/ORCA-904
**Changes:**
- Updated the recreation_fee schema to include the new fee_subcode fields.
- Added a database migration to handle incoming fee_codes from FTA and map/store them as fee_code and sub_codes into RST.
- Updated data fixtures to align with the new schema changes.
- Enhanced the API endpoints to successfully process and store both fee_codes and sub_codes.
- Updated the user interface dropdowns to properly list and handle sub_codes.
- Updated the fee display table to show both the primary fee codes and their corresponding sub-codes.

**How to Test:**
- Log into the RecSpace local environment at http://localhost:3001.
- Navigate to a specific recreation resource's fee section (e.g., http://localhost:3001/rec-resource/REC204117/fees).
- Click Add a fee.
- Verify that the Fee Type dropdown correctly displays both the fee types and their respective sub-types.


admin/backend/src/recreation-resources/fees/dto/create-recreation-fee.dto.ts
```diff
@@ -20,6 +20,20 @@ export class CreateRecreationFeeDto {
...
+    example: 'CAMPING',
+    required: false,
+  })
+  @IsString()
+  @Matches(/^[A-Z_]{1,30}$/, {
+    message:
+      'recreation_fee_sub_code must be uppercase letters/underscores (1-30 chars)',
+  })
+  @IsOptional()
+  recreation_fee_sub_code?: string;
```
diego-oxd: Shouldn't this be required when we're creating new ones? I know that the backend should allow for this to be null to accomodate previous data, but when creating new ones, shouldn't this be required?

admin/frontend/src/services/hooks/recreation-resource-admin/useCreateFee.ts
```diff
@@ -17,6 +17,7 @@ import { mapRecreationFee } from './helpers';
 export interface CreateFeeRequest {
   recResourceId: string;
   recreation_fee_code: string;
+  recreation_fee_sub_code?: string;
```
diego-oxd: shouldn't this be required?

migrations/rst/sql/V1.1.72__recreation_fee_sub_types.sql
```diff
@@ -0,0 +1,92 @@
...
+
+drop index if exists rst.recreation_fee_unique_idx;
+
+create unique index recreation_fee_unique_idx
+on rst.recreation_fee (
+    rec_resource_id,
+    recreation_fee_code,
+    recreation_fee_sub_code
+)
+where is_deleted = false and recreation_fee_sub_code is not null;
```
diego-oxd, line 124: Do we have to check on the different environments if there are existing entries that have the same type for the same resource? More broadly, since this is a new constraint, would they be dropped if they existed?
Feel free to disregard if you already checked that we don't have any such cases in the existing data in prod/test.

diego-oxd (commented):
Just quick mention so reviewers are aware: I know there is a ticket specifically for the public frontend part. But if this get merged before that. The logic for determining `recreation_fee` and `additional_fees` in `public/backend/src/recreation-resource/utils/formatRecreationResourceDetailResults.ts` will be broken until then. So the data shown will be inaccurate for that window?

But great job. It's looking really nice.

ayeshmcg:
@diego-oxd  This PR has been updated

migrations/fta/sql/V1.0.8__fta_to_rst_fee.sql
```diff
@@ -1,92 +1,3 @@
...
-           sunday_ind = excluded.sunday_ind,
-           updated_at = excluded.updated_at,
-           updated_by = excluded.updated_by,
-           is_deleted = false,
-           deleted_at = null,
-           deleted_by = null
-       where rst.recreation_fee.is_deleted = false;  -- leave deleted rows alone
-
-
--- If a fee row is removed in FTA, soft delete it in RST
```
diego-oxd, line 89: Was it part of this ticket to not import the fees data from FTA anymore? 
Or do we just need to update the mapping for the codes?
This can be tricky with the way we check for the conflict when the subcode can be null, but I think something like this might work
```
on conflict (
    rec_resource_id, recreation_fee_code,
    coalesce(recreation_fee_sub_code, ''),
    monday_ind, tuesday_ind, wednesday_ind, thursday_ind, friday_ind, saturday_ind, sunday_ind
) where is_deleted = false
``` 

and then we'd need to update the index in `migrations/rst/sql/V1.1.72__recreation_fee_sub_types.sql` to also be something like 
```
create unique index recreation_fee_unique_idx
on rst.recreation_fee (
    rec_resource_id,
    recreation_fee_code,
    coalesce(recreation_fee_sub_code, ''),
    monday_ind, tuesday_ind, wednesday_ind, thursday_ind, friday_ind, saturday_ind, sunday_ind
)
where is_deleted = false;
```

open to your thoughts

ayeshmcg: Hey @diego-oxd  Thanks for your review, I was working on testing migrations in dev, Now have the updated ones in the code, which are working in dev and the data is being populated correctly

migrations/fta/sql/V1.0.8__fta_to_rst_fee.sql
```diff
@@ -1,109 +1,182 @@
...
+    and case rf.recreation_fee_code
+            when 'C' then 'O' when 'H' then 'O'
+            when 'D' then 'A' when 'P' then 'A'
+            when 'T' then 'T'
+        end = ft.recreation_fee_code
+    and case rf.recreation_fee_code
+            when 'C' then 'C' when 'H' then 'H'
+            when 'D' then 'D' when 'P' then 'P'
+            when 'T' then null
+        end is not distinct from ft.recreation_fee_sub_code
```
diego-oxd: This can have a fan-out issue. fta_transformed correctly deduplicates FTA rows. But then the INSERT joins back to the raw fta.recreation_fee table to grab `entry_timestamp ` and `entry_userid`
That JOIN matches on resource + code, but not on day indicators. 
So for each deduped fta_transformed row, you can get multiple raw FTA rows back, and the INSERT probably will try to write all of them. Which would eventually throw an error because of the index?

I think we can just add something like `entry_timestamp as created_at` and `entry_userid as created_by` to the temp table SELECT. Then the INSERT can pull those from ft directly Then, we can remove this JOIN to fta.recreation_fee entirely.

diego-oxd (commented):
I really like the approach of using the deduplicated temp table. Great job. 
Just had that comment for the insert new records.

migrations/fta/sql/V1.0.8__fta_to_rst_fee.sql
ayeshmcg: Hey , 
Thanks for pointing that out, looks like some omissions during testing

diego-oxd (approved):

PR #1677, opened 2026-05-29
chore: created advisories table

fbarreta:
https://bcparksdigital.atlassian.net/browse/ORCA-761

Advisory table creation.


admin/backend/prisma/schema.prisma
```diff
@@ -1293,6 +1294,72 @@ model recreation_resource_status_code_history {
...
+  modified_date               DateTime                 @db.Timestamptz(6)
+  published_at                DateTime?                @db.Timestamptz(6)
+  sys_period                  Unsupported("tstzrange") @default(dbgenerated("tstzrange(CURRENT_TIMESTAMP, NULL::timestamp with time zone)"))
+  recreation_resource         recreation_resource      @relation(fields: [rec_resource_id], references: [rec_resource_id], onDelete: NoAction, onUpdate: NoAction)
+
+  @@ignore
+  @@schema("rst")
+}
+
+/// The underlying table does not contain a valid unique identifier and can therefore currently not be handled by Prisma Client.
```
ayeshmcg, line 1330: @fbarreta  
question: Will we be needing a history table as well? As we discussed the history will be captures in the staff tool.

ayeshmcg (changes requested):

admin/backend/prisma/schema.prisma
fbarreta: That's a good point, I thought about it, but it would be good as our sanity check, since we will have API calls changing our data, we could keep track of it. But if you think it's not necessary, I can remove it.

ayeshmcg: @fbarreta ACT keeps their own history/historical changes, unless they give us an api to get historical changes for a rec site we can keep the history table. So I think we can keep it

diego-oxd (approved):
Lgtm.

ayeshmcg (approved):

PR #1679, opened 2026-06-02
fix: ORCA-12 new layers proxy url

diego-oxd:
Found yet another bug with the map layers. In some situations, the requests were too large and the backend couldn't process them. We were unintentionally fetching boundaries and trail lines for over 1K features at the same time. 

Made changes so that we only fetch exactly the IDS of the resources currently in the viewport and only when we get to the right zoom for the layers to appear. 

If you can test it using the dev DB that would be ideal since the previous only surfaced after a certain amount of resources were present on the screen.


fbarreta (approved):

public/frontend/src/components/search-map/hooks/useRecreationBoundaryLayer.ts
```diff
@@ -11,24 +16,37 @@ import {
...
-    createSource,
+    createRecreationBoundarySource,
     createRecreationBoundaryLayer,
     createRecreationBoundaryStyle,
     layerOptions,
   );
+
+  const fetchByIds = useCallback(
+    (ids: string[]): Promise<Feature[]> =>
+      fetchBcgwFeaturesByIds({ url: BCGW_PROXY_URL, layer: '5', ids }),
```
ayeshmcg: Just a nit bit , it would be better to extract hardcoded layer IDs to constants
and use like 
export const BCGW_RECREATION_BOUNDARY_LAYER = '5';

public/frontend/src/components/search-map/hooks/useRecreationTrailLayer.ts
```diff
@@ -11,24 +16,37 @@ import {
...
-    createSource,
+    createRecreationTrailSource,
     createRecreationTrailLayer,
     createRecreationTrailStyle,
     layerOptions,
   );
+
+  const fetchByIds = useCallback(
+    (ids: string[]): Promise<Feature[]> =>
+      fetchBcgwFeaturesByIds({ url: BCGW_PROXY_URL, layer: '3', ids }),
```
ayeshmcg: Just a nit bit , it would be better to extract hardcoded layer IDs to constants
and use like export const BCGW_RECREATION_TRAIL_LAYER = '3';

ayeshmcg (changes requested):

ayeshmcg:
@diego-oxd  not a high priority 
Browser warning appears during hover detection
_**Canvas2D: Multiple readback operations using getImageData are faster 
with the willReadFrequently attribute set to true.**_
Cause: The forEachFeatureAtPixel() call on hover reads canvas pixels frequently

ayeshmcg (approved):

PR #1684, opened 2026-06-04
feat: ORCA-930 advisories on cards for public app

diego-oxd:
## Add advisory count and access status to resource cards
Shows advisory info on the search result and map pin cards. 

Backend:
- Added a LEFT JOIN on rst.act_advisories_flat to the search query to aggregate advisory data per resource
- Two new fields on the search response: 
   - advisory_count (number of active advisories -- normalized to 0 when null)
   -  top_access_status_grouplabel (highest-priority advisory label)

Frontend:
- New StatusIcons.tsx: inline SVG components (BlueStatusIcon, RedStatusIcon, YellowStatusIcon) to replace the old  approach, easier to extend for future status codes
- New AccessStatus.tsx: renders the status icon, description, and an advisory link ("Check advisories (N)") when advisory_count > 0
- RecResourceCard: uses AccessStatus instead of the old Status component
- Map pin popup (RecreationFeaturePreview) now also uses the card
- Added RecResourceCardResource interface so both the search model and the detail model can feed into RecResourceCard without type gymnastics

Status still comes from recreation_status.status_code: same source as before. We will update this once all the integration is fully done.


public/frontend/src/components/rec-resource/card/AccessStatus.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  if (!icon) return null;
+
+  return (
+    <div className="access-status-icon">
+      {icon}
+      {statusDescription}
+      {!hideComma && advisoryCount > 0 && ', '}
+      {advisoryCount > 0 && (
+        <a href={`/resource/${slug}#advisories`}>
```
fbarreta: The advisories are into the know-before-you-go session.
So just change the bookmark to #know-before-you-go

fbarreta (changes requested):

public/frontend/src/components/rec-resource/card/AccessStatus.tsx
diego-oxd: Ok, nice. Thanks, I pushed the change.

fbarreta (approved):
Great work ! :)

PR #1685, opened 2026-06-04
chore: implement advisories on resource page

fbarreta:
https://bcparksdigital.atlassian.net/browse/ORCA-763

Implemented frontend to display the advisories on the Status header and on Know before you go session

How to test:

Check http://localhost:3000/resource/REC204117 with 4 advisories.
Click on the header, and it should redirect to Know Before you go session.


public/frontend/src/components/rec-resource/Status.tsx
```diff
@@ -25,8 +26,14 @@ const Status = ({ description, statusCode }: StatusProps) => {
...
         width={24}
-      />{' '}
+      />
       <span>{description}</span>
+      {advisoriesCount > 0 && (
+        <>
+          {','}&nbsp;
+          <a href="#know-before-you-go">check advisories ({advisoriesCount})</a>
+        </>
+      )}
```
diego-oxd, line 38: I think this should be underlined and have a period at the end. Like in the cards. If you get the latest changes in main, you can just delete this component and reuse the `AccessStatus` which handles the styling (copying BC parks) and status.

public/frontend/src/components/rec-resource/section/AdvisoryItem.tsx
```diff
@@ -0,0 +1,82 @@
...
+  isLast,
+  isCollapsed,
+  onToggle,
+}) => {
+  return (
+    <div className={`advisory-item ${isLast ? 'advisory-item--last' : ''}`}>
+      <div
+        className={`advisory-header ${isCollapsed ? 'collapsed' : ''}`}
+        onClick={onToggle}
+        role="button"
```
diego-oxd, line 40: If we add this here, we can support keyboard nav for each item. 
```
tabIndex={0}
aria-expanded={!isCollapsed}
```

diego-oxd (commented):
Looks great. Just a couple of small issues. 

Some other general things that I will probably flag, even though I don't know what the actual expected implementation is.
- We have some flags in the model like `is_advisory_date_displayed`, `is_effective_date_displayed`, `is_end_date_displayed`, and `is_updated_date_displayed`.  So, I'm wondering this should be used to hide some of the dates to the public?
- I think your design is consistent with the design on Figma, but looks a little different than the component in BC Parks. We might have to double check if we want to keep it closer to theirs. Like the padding in the inside section or the icon size

diego-oxd (approved):
Looks good, great job!

PR #1688, opened 2026-06-09
feat: add fee sub type public site

ayeshmcg:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-916
Changes:-

- Updated schema, api to get updated fees data
- Updated frontend to have new Fee section 

To test:-

- Go to http://localhost:3001
- Select any rec resouce and add fees(type + subtype)
- Go to http://localhost:3000/resource/{rec_id}
- Ensure you see the update fee section

Note: Also udpated Rec resource page header as it was dependant on the fee


fbarreta (approved):

diego-oxd (changes requested):
Looks really good so far. 

The only things I noticed is that for the updated camping section (now overnight). The ticket mentions:
> The same logic as before applies so that this section only shows up if there are any campsites or huts or cabins (even if there are no fees, it will still show)
This section will show the number of campsites (same as before)

Like this image below. But I don't think this is being shown anymore.
Image

I believe this happens for any camping component that has no fees. 

Also, don't know if it's related but for camping components with no fees, we see the fee link in the left, but the section doesn't appear in the content. 

Image


3. I think the `Discounted rates` row is also missing in the overnight fees.
4. feeTypeMap probably needs to be updated to have the new types.

ayeshmcg:
@diego-oxd  this has been updated , also regarding Discounted rates those dont need to be included now as discussed with Airlia and Alex
this pr also includes updated rec resourc header

diego-oxd (approved):
Yeah, looks great now!

diego-oxd (approved):

PR #1690, opened 2026-06-11
feat: ORCA-932 public access status column

diego-oxd:
Don't merge until https://github.com/bcgov/nr-rec-resources/pull/1687 has been merged.

## Backend changes
- Added `public_access_status` as a filter param to the search endpoint. The filter matches resources based on their top ranked advisory only (using the same sort order provided by team Lynx), same order used elsewhere in the app. Resources with no advisories default to "Open".
- The search result row now includes `access_status_grouplabel`, pulled from `act_advisories_flat` using the same ranking with take: 1 (so the displayed value is always the top-ranked advisory).

## Frontend changes
- Added a "Public access status" column to the search results table (feature-flagged). It displays a colour coded badge using the access_status_grouplabel value from the API. The colour mapping is: grey for Closed/Restricted, yellow/gold for Visit with caution/Limited access, and green for Open/Seasonal restrictions. If the field is null (no advisories), it defaults to showing "Open" with the green badge.
- Added a "Public access status" filter to the filter panel (also feature-flagged). The filter options are hardcoded to the known group label values (Open, Seasonal restrictions, Visit with caution, Limited access, Restricted, Closed).

### DIFF
--- a/admin/backend/src/recreation-resources/dtos/admin-search-query.dto.ts
+++ b/admin/backend/src/recreation-resources/dtos/admin-search-query.dto.ts
@@ -36,6 +36,8 @@ export const ADMIN_SEARCH_SORT_VALUES = [
   'district:desc',
   'display_on_public_site:asc',
   'display_on_public_site:desc',
+  'public_access_status:asc',
+  'public_access_status:desc',
 ] as const;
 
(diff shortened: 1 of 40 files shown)


fbarreta (approved):
Look great !

PR #1694, opened 2026-06-22
feat: added super admin role

ayeshmcg:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-957
**What changed**
**Search table** -  archived resources are now visible to all admins when browsing (no text query). When a text search is active, only rst-super-admin sees archived results.
**Typeahead suggestions** - archived resources are no longer suggested to non-super admins, keeping suggestions consistent with search results.
**Edit guard** - mirrors the frontend EditableGuard on the API: non-super admins receive 403 Forbidden when attempting to update an archived resource.
**New auth utilities (reusable for future features)**
**@IsSuperAdmin() -** parameter decorator that injects a boolean from the JWT token directly into any handler.
**SuperAdminGuard -** a NestJS guard to lock an entire endpoint to rst-super-admin only via @UseGuards(SuperAdminGuard).

### DIFF
--- a/admin/backend/src/auth/auth.constants.ts
+++ b/admin/backend/src/auth/auth.constants.ts
@@ -7,6 +7,7 @@ export const AUTH_ROLES_KEY = 'roles';
 export enum RecreationResourceAuthRole {
   RST_ADMIN = 'rst-admin',
   RST_VIEWER = 'rst-viewer',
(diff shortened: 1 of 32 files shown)


fbarreta (approved):
Looks great ! Tks !

PR #1743, opened 2026-07-24
chore: super admins can edit resource name

fbarreta:
Cards: https://bcparksdigital.atlassian.net/browse/ORCA-949 and https://bcparksdigital.atlassian.net/browse/ORCA-1007
Allows Super Admins to edit resouce name and display overview tab as default.

- Backend check to prevent unwanted changes
- Backend dto changes to allow name change
- Display overview tab as default
- Frontend Edit Session added field
- Changed the flag from canEditArchived to isSuperAdmin so it can fit in multiple functions allowed to only super admins.


ayeshmcg (approved):
LGTM

PR #1746, opened 2026-07-27
chore: fix query multiple space search

fbarreta:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-970

Fix query so multiple spaces or single space will not affect the search.


ayeshmcg (approved):
Great work! Could we add a test for this change before merging?

PR #1747, opened 2026-07-27
fix: Use hook to display mobile view instead of just css

diego-oxd:
The issue was that `__root.tsx` was mounting `<Outlet />`, which contains the map component, twice. One for mobile and one for desktop. It was relying on css to hide/display one of them based on the screen size. However, the elements were still there. 
The `prp-map` library relies on getting the map container by a unique id `map-container`. 
Since there were 2 components with the same ID, the library was finding the first component with that ID, the mobile one (hidden) and adding all the map visual components there. So, for desktop we only had the empty map container. 

Fixed by using a hook to return only the proper components instead of pure CSS to hide/display them. Now, we will only have the correct map-container displayed.


ayeshmcg (approved):
Nice work @diego-oxd

PR #1752, opened 2026-07-29
chore: add rec_resource_id to order and highlighter

fbarreta:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-1007

This PR adds:

- highlight to rec_resource_id in case of query match
- Includes rec_resource_id on sort in case of match, so it behaves like the resource name, bringing the matches first


diego-oxd (approved):
lgtm

PR #1753, opened 2026-07-29
chore: scroll to top when switch tab

fbarreta:
Card: https://bcparksdigital.atlassian.net/browse/ORCA-1094

Move scroll to the top when switch to rec resource layout.


diego-oxd (approved):
lgtm

PR #1765, opened 2026-08-06
chore: added feedback card to map

fbarreta:
Card for feedback card: https://bcparksdigital.atlassian.net/browse/ORCA-1011
Card for matomo tracking: https://bcparksdigital.atlassian.net/browse/ORCA-1075
Added feedback card.
image


diego-oxd (approved):
Looks great man. 

Just a couple of issues I noticed. 

For mobile, the card slightly overlaps the other buttons. 

Image

This might be an issue that exist before this PR, but I don't think the calls to `trackClickEvent` are working properly. Like the function is not being invoked. 

I'll approve it anyway, so you can merge after the small fixes.

fbarreta:
> Looks great man.
> 
> Just a couple of issues I noticed.
> 
> For mobile, the card slightly overlaps the other buttons.
> 
> Image
> This might be an issue that exist before this PR, but I don't think the calls to `trackClickEvent` are working properly. Like the function is not being invoked.
> 
> I'll approve it anyway, so you can merge after the small fixes.

Thanks ! I fixed the layout, we can test on dev or test the tracking issue, and maybe investigate it better !

